PR #5, opened 2021-12-27
glutin: handle mouse scroll, set mouse state

rhelmot:
I ran into some missing features while building a view with mouse interactivity. These fix everything on my end.

Please let me know if you want me to write tests!


geom3trik:
This looks good to me, thank you for the contributions!

PR #6, opened 2021-12-31
Add radio buttons view

rhelmot:
Before I sink too much time into this, I wanted some feedback - is this an appropriate design? The aggressive use of downcast in the builder seems problematic, but I don't know how else you would accomplish my design goals (can construct arbitrary contents for each radio item; clicking on any of these items will select the corresponding index).

### DIFF
--- a/core/src/views/radio_buttons.rs
+++ b/core/src/views/radio_buttons.rs
@@ -1,23 +1,16 @@
-use std::collections::HashMap;
-use crate::{Context, Entity, Event, Handle, View, WindowEvent, MouseButton, PseudoClass, Label, LayoutType, HStack, VStack, TreeExt};
+use crate::{Context, Entity, Event, Handle, View, WindowEvent, MouseButton, PseudoClass, Label, LayoutType};
 use crate::style::PropSet;
 use crate::Units::*;
 
 pub struct RadioButtons {
-    pub idx: usize,
     on_changed: Option<Box<dyn Fn(&mut Context, usize)>>,
-    entity_map: HashMap<Entity, usize>,
-    check_boxes: Vec<Entity>,
-    icon_checked: String,
-    icon_unchecked: String,
 }
 
-const ICON_CHECK: &str = "\u{2713}";
+const ICON_CHECK: &str = "\u{00B7}";
 
 impl RadioButtons {
     pub fn with_icons<F>(
         cx: &mut Context,
-        layout: LayoutType,
         idx: usize,
         idx_count: usize,
         builder: F,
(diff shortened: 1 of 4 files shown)


geom3trik (commented):
Overall this is a good attempt at a radio buttons widget but it's perhaps more in line with the design of tuix, which has a more widget-first approach, as opposed to vizia, which is intended to be data-first.

What I mean by this is that vizia is designed to be reactive. So changes in data result in updates of the UI. So the first thing to think about when designing a widget is what data it's supposed to represent. For radio buttons that would be something like an enumerated type, though I don't know whether or not the identifiers would be part of the widget. If not then I guess it would be a list of booleans where only one can be true at a time.

So, instead of duplicating the state inside the widget as you've done, you can use the data system and bindings to bind directly to some piece of user data and update the widget accordingly. Perhaps the easiest way to do this is to take a lens as an argument to the constructor and then use a `Binding` view inside of the build closure. You can then use the `Checkbox` widget (it may need some modification because I don't think I got custom icons working yet) to reflect the state of the data.

Depending on what you use for the data this might be a bit different but if you take a look at the `Picker` widget and the corresponding `picker.rs` example you can see my attempt to create a widget for enumerated data. The picker widget declares a binding to some data, determined by the lens, and then there are separate widgets (in this case colored buttons) created by the user for each item. This approach was necessary because there's no way to tell how many variants are in the enum and also because each item needs emit a user event to update the data.

If you go for the approach of a list then you can figure out how many items there are and generate the sub-widgets (Checkboxes and Labels) automatically. But then it's more difficult to update that data with events because it's still on the user to provide the event and make sure it does the right thing (a trait could maybe be used here maybe). Alternatively, you could go the approach of the `List` widget and generate some extra data which keeps track of the 'selected' item. So the user provides the list of names and the `RadioButtons` widget keeps track of which one is checked (similar to how you have it now) but using data that's built into the data system rather than inside the widget. Take a look at `ListData` and its use in `List` to see what I mean.

I hope the feedback doesn't come across as too negative as I don't mean it that way. I'm still figuring out myself the best approach to creating some of these widgets. Thank you for your time and effort on this!

rhelmot:
Should the checkbox widget be redesigned then? I modeled this as closely as I could on that widget, and it duplicates its state too.

Additionally, the data I want to model really is just an index into a list. I'll try try modeling it more closely after the picker.

rhelmot:
I've pushed a second draft. It's a lot cleaner now!

core/src/views/radio_buttons.rs
```diff
@@ -0,0 +1,118 @@
...
+        icon_unchecked: String,
+    ) -> Handle<Self>
+    where
+        F: 'static + Fn(&mut Context, usize) + Clone
+    {
+        Self {
+            on_changed: None,
+        }
+            .build2(cx, move |cx| {
+                for idx_available in 0..idx_count {
```
geom3trik, line 27: There's a subtle quirk when you create a variable number of sub-widgets inside of the build closure. Essentially vizia uses the tree to determine if a widget should be created or just updated. Because this widget could be within a binding which changes the `idx_count`, and thus could change the number of `RadioButton` widgets created, it will cause a bug because the tree won't match. 

The easiest solution to this is to remove the children of the widget before creating any new ones, so essentially forcing it to rebuild its children rather than update them. For this reason using a `for` loop or an `if` statement to selectively build widgets doesn't work in the general case. For the for loop you can use the `ForEach` view, which does the child removal part for you.

core/src/views/radio_buttons.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn new<F>(cx: &mut Context, idx: usize, icon_text: String, builder: F) -> Handle<RadioButton>
+    where
+        F: 'static + FnOnce(&mut Context)
+    {
+        RadioButton {
+            idx
+        }.build2(cx, move |cx| {
+            let entity = cx.current;
+            entity.set_layout_type(cx, LayoutType::Row);
+            entity.set_height(cx, Stretch(0.0));
```
geom3trik, line 109: The `build2` (will be renamed soon) method returns a `Handle` which you can set properties on directly. So instead of this you can do `build2(...).layout_type(LayoutType::Row).height(Stretch(0.0)`.

Also, I think you wan't `Stretch(1.0)` and not `Stretch(0.0)` unless I'm misunderstanding what you're trying to do.

core/src/views/radio_buttons.rs
```diff
@@ -0,0 +1,118 @@
...
+    idx: usize,
+}
+
+impl View for RadioButton {
+    fn element(&self) -> Option<String> {
+        Some("radiobutton".to_owned())
+    }
+
+    fn event(&mut self, cx: &mut Context, event: &mut Event) {
+        if let Some(WindowEvent::MouseDown(btn)) = event.message.downcast() {
```
geom3trik, line 91: ooh, I didn't even know you could do this.

geom3trik (commented):
This looks really good and is closer to what I had in mind for how widgets should be made in vizia.

Referring to your previous comment about the `Checkbox` storing its own state, you are quite right. After thinking about it some more I guess roughly there are a few categories of widgets: There's widgets that contain their own state, like the checkbox and the label, but could be placed within a `Binding` view to bind their internal values to some app data; there's widgets which derive their data directly from some app data by using a binding internally and usually take a lens as an argument, like the `List` or `Picker; and then I guess you could have a widget that can do both even though I haven't made one like that yet.

I'm not sure yet what the rule is for when to use each of these options. So for this `RadioButtons` widget you've gone for the first option where it stores its own state but could be places within a binding to link it to some user data. I think this a perfectly valid approach but I could also envision a version which takes a lens to some list and binds to it directly.

I think what would be useful is an example of the intended use of this widget and how it could be used within a binding. Would you be able to add an example to the `examples/controls` directory?

Overall though this looks really good and I'd be happy to merge this with a few minor changes, but I definitely would also like to understand how it should be used as well. Thanks again for your time and effort on this!

PR #9, opened 2022-01-07
Custom knob building method

Fredemus:
This pull request adds a method to the `Knob` widget, `Knob::custom()`. 

It does the same as the `new()` methods, except instead of constructing an `ArcTrack()` it takes a closure which the user can use to build whatever thing to represent the knob that they want. 

For example, this just builds a text field showing the "knob" value, which can be dragged to change it like a knob:

```rust
Knob::custom(
            cx,
            default,
            value,
            move |cx, val| {
                Label::new(cx, &format!("value: {}", val))
            }
        );
```

### DIFF
--- a/core/src/views/knob.rs
+++ b/core/src/views/knob.rs
@@ -74,6 +74,43 @@ impl Knob {
             });
         })
     }
+    pub fn custom<'a, F>(
+        cx: &'a mut Context,
+        normalized_default: f32,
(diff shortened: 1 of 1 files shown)


geom3trik (approved):
This looks good to me. I'm going to merge this as is but there will probably be some minor changes when #10 gets merged. Oh and also to make everything configurable from the knob, like radius and span.

PR #11, opened 2022-01-08
Add radio button knob widget

rhelmot:
Supersedes #6 

Much simpler concept. This time, it's just a clone of the checkbox widget with custom drawing logic.

Also makes propget take &Context instead of &mut Context so it can be used within draw.


core/src/views/radio_buttons.rs
```diff
@@ -0,0 +1,78 @@
...
+                (f)(cx);
+                self.on_select = Some(f);
+            }
+        }
+    }
+
+    fn draw(&self, cx: &Context, canvas: &mut Canvas) {
+        let entity = cx.current;
+        let bounds = cx.cache.get_bounds(entity);
+        let border_width = if bounds.w < bounds.h { bounds.w } else { bounds.h } / 6.0;
```
geom3trik, line 35: Where does this come from? Is this just what makes it look correct? I only ask because there is a border width style property already which can be specified by the user. My thinking is that it would be good for the user to be able to configure the border width as well as the size (radius?) of the inner circle.

geom3trik (commented):
This looks good! Just a quick query about border width and then I'm happy to merge this.

core/src/views/radio_buttons.rs
rhelmot: I just pushed a fix to use the border-width property. I initially didn't use it because I couldn't come up with a good way to make the inner dot stylable by the user, but I ended up just taking half of the leftover space after drawing the border and allocating it to the dot.

PR #15, opened 2022-01-15
Static lenses and list callbacks

rhelmot:
Supercedes #13 

This PR is a tragic victim of feature creep. Feel free to decide that you want any subset of the features here.

- Adds StaticLens, which lets you bind to any static data instead of just data in a model.
- Adds an example of using StaticLens with a list in the selectable_list example. Since there are now two lists, this breaks the usage of ListEvent::IncrementSelection and DecrementSelection.
- Fix increment and decrement in the example by adding callbacks to the list view that are triggered at the same time as the increment and decrement events.
- Remove ListData and ListEvent, since they are now totally superceded by the callbacks. Additionally, this adds one more callback, a clear callback which is triggered when the list receives an escape keypress. This makes the multi-selection list example actually function.


geom3trik (commented):
I like all of these additions. The one thing I would change though is to move the logic which increments/decrements the selected index into the model, that way you don't need to do `cx.data()` inside of the increment/decrement callbacks and you don't need to wrap the whole list in a binding to the selected index. What are your thoughts on that?

I don't think it makes a difference in this case, because the list only updates if the number of elements changes, but having the list inside the binding for the selected index will cause the list to update whenever the selected index changes. This might make a difference if the number of items does change it might cause the list to get updated twice. In this simple example it doesn't actually do anything, but depending on what's in the list it might matter.

You make a good point that the ListData and ListEvent aren't really needed. But if the logic for selection should be in the model, it does raise the question whether there should be *something* which makes this easier for the user. So basically the options are:

1. Let the user handle the list data and selection logic (most flexible but mor code for user to write)
2. Provide something, like `ListData` which the user can include in there own model and which handles the logic for them (could be tricky for multiple lists).
3. Let the list handle its own list data (then need a way to propagate any changes to the users model data if they need to track the selection themselves, possibily duplicating data).

rhelmot:
Updated as per your comments on incrementing and decrementing. Interestingly, this uncovered another bug in list - previously, list was only ever set to focused on build, which worked when it was either a) the only list or b) rebuilt on every change, but when we're not doing that it does not work. I added some logic to focus the list on click if none of its children have grabbed focus first.

As much as I really want to simplify the user-logic around increment and decrement events, I don't think it's reasonable to implement unless the list owns (explicitly builds a model containing) its own data. List is pretty simple now, do we maybe want to have a second list which functions that way?

geom3trik:
Discussed on discord but repeating here for convenience: I'm going to merge this as is for now and then work on a separate list view which tracks its own selection state, which may be possible to merge into the simple `List` view.

PR #18, opened 2022-01-19
Make lenses fallible

rhelmot:
As outlined on discord, this makes it so that Lens::view returns Option<'&a Self::Target>. This is mostly abstracted away by adding an unwrap() in Field::get. This can be interacted with safely by using Binding::new_fallible, which has a TODO which I will need some help resolving.

### DIFF
--- a/core/src/state/binding.rs
+++ b/core/src/state/binding.rs
@@ -19,6 +19,10 @@ where
     builder: Option<Box<dyn Fn(&mut Context, Field<L>)>>,
 }
 
+struct FallibleMarker(bool, usize);
+
+impl Model for FallibleMarker {}
+
 impl<L> Binding<L>
 where
     L: 'static + Lens,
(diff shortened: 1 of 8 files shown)


rhelmot:
If we like this and we think that vizia is young enough to make such a big breaking change, I would like to propose that Binding::new and Binding::new_fallible be renamed to Binding::new_unwrap and Binding::new.

rhelmot:
Good news! I figured out how to address the TODO I previously mentioned. This PR is now ready for merging, assuming you like the API.

As a bonus, I made Lens::index work. I changed the commented implementation a bit because using Then required the intermediate data to be Model, which was undesirable in my use case.

rhelmot:
The two subtle bugs mentioned in the above commit message are:

a) if a fallible binding had the initial value of none, it would fail to cache any data. I resolved this by adding the `force` parameter.
b) if fallible bindings are directly nested inside each other, they would override each other's cached data. I resolved this by forcing the system to cache on the most recent child, i.e. the binding itself.

rhelmot:
Superseded by #33 and #41

PR #23, opened 2022-01-24
Print warning on unknown rules, do not rebuild application on f5

rhelmot:
Two changes which I found helpful while debugging my application.

- first change is self-explanitory
- second change is specifically motivated by a) being annoyed that my state would reset whenever I hit f5 (the element I was trying to style is not visible by default in my app) and b) noticing that once I added the first change, messages would appear a quadratically increasing number of times each time I hit f5 - each rebuild loads a new theme without clearing the old themes. I don't know if you also want to add logic clearing the user themes and stylesheets to the application logic that rebuilds everything, but this seemed like the easiest solution.

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -236,7 +236,7 @@ impl Context {
 
         self.style.parse_theme(&overall_theme);
 
-        self.enviroment.needs_rebuild = true;
+        // self.enviroment.needs_rebuild = true;
 
         // Entity::root().restyle(self);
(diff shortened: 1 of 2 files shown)


geom3trik:
I have a feeling that not rebuilding when reloading might cause some problems, but after testing this with my own app it doesn't seem to cause any noticeable problems so time will tell. As for the styles not getting properly replaced that was an oversight. I'll merge this for now as a solution but at some point I would like to revisit this problem.

Issue #25, opened 2022-01-30
Persistence of View/Model data


geom3trik:
Currently when a view is built using `Self{ }.build2(cx, |cx|{ })` the closure is run every time but the view is not added to the context if it already exists. This means that any changes to fields in `Self { }` are not updated when the view is rebuilt. 

This isn't necessarily a problem because any data that did need to be updated could live in a model which *is* updated when rebuilt. So when you have some data `AppData{ }.build(cx)`, this is updated with new values every time it is called, so calling it inside a binding updates that data every time the binding is rebuilt. 

This whole system presents a few problems. The first is that there are two different places data can live and two different behaviours that can happen when rebuilt:

1. Data inside a view struct, **not** updated when rebuilt
2. Data inside a view struct, updated when rebuilt
3. Data inside a model, **not** updated when rebuilt
4. Data inside a model, updated when rebuilt

Data inside of views is local to the view wheras data inside of models can be shared accross multiple views.

The second problem is that even when data is updated there is no easy access to the previous version of the data without having to get a view/model from context and downcasting it to the right type. This becomes a problem if you want to conditionally update a view/model. A further problem is that sometimes it's necessary to persist model data, resulting in an ugly `if cx.data::<AppData>().is_none()` check.

I think what is needed is an explicit way for the user to determine if the view/model data should be updated or not if there is already a version in context. I'm not sure of the exact systax yet, but perhaps something like:

```rust
cx.build(|cx, prev|{
    if let Some(previous_data) = &prev {
        // Choose to not update or could return Some(Self{ }) which would update it
        return None;
    } else {
        // Create a new one if there's no previous one
        return Some(Self{ }); 
})
```
This isn't ideal because it shoudn't be possible to *not* create a new instance when there isn't a previous one but it's a start. I'm very much open to suggestions.

rhelmot:
How about this for cx.build?

```
fn build<T, FN: FnOnce(&mut Context) -> T, FU: FnOnce(&mut Context, &mut T)>(&mut self, new_builder: FN, update_builder: FU)
```

That way, we call new_builder if the data is being constructed for the first time and then insert the return value into the state, and we call update_builder with a mutable reference to the value already in the state if one exists.

geom3trik:
> ```
> fn build<T, FN: FnOnce(&mut Context) -> T, FU: FnOnce(&mut Context, &mut T)>(&mut self, new_builder: FN, update_builder: FU)
> ```

[...]

PR #28, opened 2022-02-04
First draft of rearrangable items

rhelmot:
This adds a new widget `Rearrangable` which allows its children to be dragged and reordered. It has several problems:

1) It requires its closures to be Clone, and then makes judicious use of this trait. I'm not sure how else to code this to make the lifetimes happy.
2) You can currently only drag a widget one space at a time, then you have to grab it again. I suspect this is because of #25 causing the holding state to be reset whenever the data updates.

The main use case I have in mind for this is to build tabs that can be reordered.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,10 @@ path = "examples/lists/multiselectable_list.rs"
 name = "static_list"
 path = "examples/lists/static_list.rs"
 
+[[example]]
+name = "rearrangable"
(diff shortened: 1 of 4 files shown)


geom3trik:
I don't know if it makes much difference in performance, but when I've needed to pass closures into bindings as you have here I've first wrapped them in `Rc` which removes the need for the closure to be clone but still allows you to clone the `Rc` and move it into the build closure.

PR #29, opened 2022-02-05
Add Context::spawn for spawning threads that can emit events

rhelmot:
Sample usage is:

```rust
cx.spawn(|cx| {
    if let Some(map) = load_workflow() {
        cx.emit(AppEvent::Load { map: RefCell::new(Some(map)) });
    }
})
```

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, VecDeque};
+use std::fmt::{Debug, Display, Formatter};
 
 #[cfg(feature = "clipboard")]
 use copypasta::ClipboardContext;
@@ -269,10 +270,24 @@ pub struct ContextProxy {
     pub event_proxy: Option<Box<dyn EventProxy>>,
 }
 
+#[derive(Debug)]
 pub enum ProxyEmitError {
     Unsupported,
     EventLoopClosed,
 }
+
+impl Display for ProxyEmitError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ProxyEmitError::Unsupported =>
+                f.write_str("The current runtime does not support proxying events"),
+            ProxyEmitError::EventLoopClosed =>
+                f.write_str("Sending an event to an event loop which has been closed")
+        }
+    }
+}
+
+impl std::error::Error for ProxyEmitError {}
 
 impl ContextProxy {
     pub fn emit<M: Message>(&mut self, message: M) -> Result<(), ProxyEmitError> {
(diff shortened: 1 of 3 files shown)


core/src/context.rs
```diff
@@ -246,4 +250,59 @@ impl Context {
...
+    }
+}
+
+/// A bundle of data representing a snapshot of the context when a thread was spawned. It supports
+/// a small subset of context operations. You will get one of these passed to you when you create a
+/// new thread with `cx.spawn()`.
+pub struct ContextProxy {
+    pub current: Entity,
+    pub event_proxy: Option<Box<dyn EventProxy>>,
+}
```
geom3trik, line 270: My only comment for this would be, what else do you think could/should go in here? Are there any properties in `Context` that could be shared atomically between threads? E.g. mouse position or current entity.

core/src/context.rs
```diff
@@ -246,4 +250,59 @@ impl Context {
...
+
+pub enum ProxyEmitError {
+    Unsupported,
+    EventLoopClosed,
+}
```
geom3trik, line 275: This should probably implement `std::error::Error`.

core/src/context.rs
```diff
@@ -246,4 +250,59 @@ impl Context {
...
+
+impl ContextProxy {
+    pub fn emit<M: Message>(&mut self, message: M) -> Result<(), ProxyEmitError> {
+        if let Some(proxy) = &self.event_proxy {
+            let event = Event::new(message)
+                .target(self.current)
+                .origin(self.current)
```
geom3trik, line 282: Since the `ContextProxy` passed to the thread is mutable the user could for whatever reason change the `current` entity before sending an event, which might cause some unexpected bugs. I don't know though if this is something that should be prevented or if that would just be considered 'user error'?

geom3trik (commented):
Looks good to me! My only question is about what should the `ContextProxy` contain, but I'm happy to merge this as is and we can figure out what else it should have in a later PR perhaps?

core/src/context.rs
rhelmot: The same thing could be said about the normal Context. Should we treat this one differently?

core/src/context.rs
rhelmot: Done.

core/src/context.rs
rhelmot: I looked at the fields and methods on Context, and couldn't find anything that seemed a) useful and b) worth the ergonomics/performance impact of Arc<Mutex<T>>.

geom3trik: If they're atomic you don't need the mutex. But yeh I think we'll leave it for now and if at some later date it becomes useful to have more things in `ContextProxy` then we can do that.

core/src/context.rs
geom3trik: Fair point. Let's leave it for now then.

PR #30, opened 2022-02-06
impl Data for PathBuf

rhelmot:


rhelmot:
Ah, whoops, I didn't see that you'd already done this. That's on me for not working with the latest main and blindly cherry-picking.

PR #39, opened 2022-02-15
Remove a few static bounds

rhelmot:
Getting rid of these bounds enables better ergonomics and fewer clones (especially in the rebuild-always branch) at no penalty.


geom3trik:
Looks good!

PR #40, opened 2022-02-15
Move Model trait bound off of Lens declaration

rhelmot:
When you do Data::one.then(One::two), you don't need One to be a model.


geom3trik:
In the 7GUIs branch (which I should merge because it has the lens map changes) I removed the `Model` bound entirely. Do you think the bound is needed on the binding source as a constraint or could we get away without it?

rhelmot:
Oh, yes, that will work much better.

PR #44, opened 2022-02-25
Add toggle_class methods to add or remove a class

rhelmot:
This adds logic for classes to be removed, instead of just added. Additionally, class membership can be bound to a bool. I updated the number_input example to demonstrate this.

### DIFF
--- a/core/src/handle.rs
+++ b/core/src/handle.rs
@@ -78,6 +78,20 @@ impl<'a, T> Handle<'a, T> {
         self
     }
 
+    pub fn toggle_class(self, name: &str, applied: bool) -> Self {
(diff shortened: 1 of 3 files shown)


rhelmot:
I would actually prefer if we could make the breaking change of replacing the existing `class` methods with these.

geom3trik:
I think this is a great addition. I'm not sure about completely replacing the `class` method but if we did I think we should refer to what JS does for this. Apparently the modern way to do this is with `classList`: <https://developer.mozilla.org/en-US/docs/Web/API/Element/classList>, which in practise would mean having something like:

- `add_class()`
- `remove_class()`
- `replace_class()`
- `toggle_class()`

PR #76, opened 2022-03-16
Property Animations

geom3trik:
Adds support for animating style and layout properties.

## Example
```rust
let animation = cx
    .add_animation(std::time::Duration::from_secs(1))
    .add_keyframe(0.0, |keyframe| keyframe.set_background_color(Color::red()))
    .add_keyframe(1.0, |keyframe| keyframe.set_background_color(Color::blue()))
    .build();

Element::new(cx)
    .size(Pixels(200.0))
    .background_color(Color::red())
    .on_press(move |cx| cx.play_animation(animation));
```

Animations can also be applied as CSS transitions:
```css
.test {
    background-color: red;
    transition: background-color 2.0 0.0;
}

.test:hover {
    background-color: blue;
    transition: background-color 2.0 0.0;
}
```


rhelmot:
My three main concerns are:

1) the use of polling to animate. I believe this is excessive and we should have a timer to control animations, pending the timers PR.
2) we're accumulating a lot of code which is just doing the same thing for each style property. We should maybe consider writing a for_each_style function or macro, or maybe like for_each_animatable_style.
3) We still need to figure out how to animate something disappearing

Other than that this seems good, and I think these concerns can be deferred to other issues.

PR #83, opened 2022-03-18
Clippy fix

geom3trik:
Fixes most clippy warnings


rhelmot:
This should also probably include adding a clippy pass to CI

geom3trik:
This is now very out of date and it was broken anyway. Going to close thise for now and tackle clippy fixes another time.

Issue #81, opened 2022-03-18
30pt text as 2x DPI scaling renders badly


robbert-vdh:
We discussed this on the Discord, but I'll also post it here so it doesn't get forgotten about. This is what happens when you render a label with a 30 point font size and the Noto Sans Thin font and 2x scaling (`application.with_scale_factor(WindowScalePolicy::ScaleFactor(2.0)` with #80):

![afbeelding](https://user-images.githubusercontent.com/748520/159001547-3f6878a1-97e0-4f6a-9e68-37c63633c164.png)

The other text, which is 15pt Noto Sans Light, looks fine. Let me know if you need more details for reproducing this.

geom3trik:
Can you confirm if this is fixed by #84?

robbert-vdh:
The text now renders correctly here at all scaling levels.

PR #84, opened 2022-03-19
Fix DPI scaling

geom3trik:
Apply DPI scaling in vizia rather than relying on femtovg

### DIFF
--- a/baseview/src/application.rs
+++ b/baseview/src/application.rs
@@ -306,8 +306,7 @@ impl ApplicationRunner {
     }
 
     pub fn render(&mut self) {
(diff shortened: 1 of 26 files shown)


robbert-vdh:
(also posting this here for archiving purposes)

Didn't seem to make any difference for the example from #81 at least:

afbeelding

And one potential problem with having to manually multiply font size by a DPI scale is that it's probably very easy to forget to do that when creating widgets, and you'd only notice that when someone tires to use your GUI on a HiDPI screen.

rhelmot (approved):

PR #99, opened 2022-04-03
Add ability to bind to local view data

geom3trik:
This allows descendant views to bins to data in their anscestors:
```rust
use vizia::*;

#[derive(Lens)]
pub struct CustomView {
    some_data: String,
}

impl CustomView {
    pub fn new<'a>(cx: &'a mut Context, txt: &str) -> Handle<'a, Self> {
        Self {
            some_data: txt.to_owned(),
        }.build2(cx, |cx|{
            Label::new(cx, CustomView::some_data)
                .on_press(|cx| cx.emit(CustomEvent::ChangeData));
        })
    }
}

pub enum CustomEvent {
    ChangeData,
}

impl View for CustomView {
    fn event(&mut self, _: &mut Context, event: &mut Event) {
        if let Some(custom_event) = event.message.downcast() {
            match custom_event {
                CustomEvent::ChangeData => {
                    self.some_data = "Something".to_owned();
                }
            }
        }
    }
}

fn main() {
    Application::new(WindowDescription::new(), |cx|{
        CustomView::new(cx, "Test");
    }).run();
}
```
In the above the `Label` binds to the local data within the `CustomView`.

### DIFF
--- a/core/src/events/event_handler.rs
+++ b/core/src/events/event_handler.rs
@@ -16,7 +16,7 @@ pub trait ViewHandler: Any {
 
 impl dyn ViewHandler {
     /// Check if a view handler is a certain type.
-    pub fn is<T: Any + 'static>(&self) -> bool {
+    pub fn is<T: Any>(&self) -> bool {
         // Get TypeId of the type this function is instantiated with
         let t = TypeId::of::<T>();
 
--- a/core/src/events/event_handler.rs
+++ b/core/src/events/event_handler.rs
@@ -42,7 +42,7 @@ impl dyn ViewHandler {
     /// Attempt to cast a view handler to an immutable reference to the specified type.
     pub fn downcast_ref<T>(&self) -> Option<&T>
     where
-        T: Any + 'static,
+        T: Any,
     {
         if self.is::<T>() {
             unsafe { Some(&*(self as *const dyn ViewHandler as *const T)) }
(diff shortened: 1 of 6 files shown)


core/src/events/event_handler.rs
```diff
@@ -16,7 +16,7 @@ pub trait ViewHandler: Any {
 
 impl dyn ViewHandler {
     /// Check if a view handler is a certain type.
-    pub fn is<T: ViewHandler + 'static>(&self) -> bool {
+    pub fn is<T: Any + 'static>(&self) -> bool {
```
KDecay, line 19: I'm pretty sure you can remove the `'static` constraint, because `Any` already implies it.

```suggestion
    pub fn is<T: Any>(&self) -> bool {
```

core/src/events/event_handler.rs
```diff
@@ -30,7 +30,7 @@ impl dyn ViewHandler {
     /// Attempt to cast a view handler to a mutable reference to the specified type.
     pub fn downcast_mut<T>(&mut self) -> Option<&mut T>
     where
-        T: ViewHandler + 'static,
+        T: Any + 'static,
```
KDecay, line 33:
```suggestion
        T: Any,
```

core/src/events/event_handler.rs
```diff
@@ -42,7 +42,7 @@ impl dyn ViewHandler {
     /// Attempt to cast a view handler to an immutable reference to the specified type.
     pub fn downcast_ref<T>(&self) -> Option<&T>
     where
-        T: ViewHandler + 'static,
+        T: Any + 'static,
```
KDecay, line 45:
```suggestion
        T: Any,
```

rhelmot:
This seems pretty straightforwardly good!

PR #102, opened 2022-04-04
Simplify slider view

geom3trik:
This PR simplifies the slider control, making use of the local data binding introduced in #99. The changes also include removing the need to pass an `Orientation` to the slider, instead deriving it from the slider dimensions after layout.

Another thing we should consider as a possible addition to this PR is whether or not the slider should take a specified `min`/`max` instead of  working only with a normalized value.

### DIFF
--- a/core/src/default_theme.css
+++ b/core/src/default_theme.css
@@ -271,11 +271,9 @@ slider.vertical {
     top: auto;
     bottom: auto;
     height: 1s;
-    width: 10px;
+    width: 5px;
 }
 
-slider track {
-}
 
 slider .active {
     background-color: #005a9e;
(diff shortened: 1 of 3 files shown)


rhelmot (approved):
Looks good. I think we should add the ability to set the slider domain so you're not always picking from a normalized range - that would reduce the number of map lenses you would have to use.

geom3trik:
> I think we should add the ability to set the slider domain so you're not always picking from a normalized range - that would reduce the number of map lenses you would have to use.

I agree. My initial thinking is to use `std::ops::Range<f32>`. The question is whether to have this as an argument or as a modifier on the handle?

Also, should we have an option for steps to make it a stepped slider? Or should that be a separate control?

rhelmot:
I think stepping is necessary, but from what I can tell, you can only do a stepped knob via the custom method. Do we want to make stepping more easily accessible from both slider and knob, or do we want to add a slider custom constructor?

geom3trik:
> Do we want to make stepping more easily accessible from both slider and knob, or do we want to add a slider custom constructor?

I'll take a look at simplifying the knob as well. I think my preferred API for making both the knob and slider stepped is through a view modifier:

```rust
Slider::new(cx, lens, range).steps(...);
```

But I need to check if this is viable. The custom function on the knob allows for a lot more customizability than just adding stepping.

rhelmot (approved):

PR #103, opened 2022-04-04
Add `Keymap`

KDecay:
I added a basic `Keymap` implementation with two examples. This should definitely be expanded to also be able to read keybindings from files using a `from_file` method, but I didn't want to add everything all at once.

Currently the `Keymap` supports inserting and removing keybindings, both during runtime and during the creation of the model.

### DIFF
--- a/core/src/input/binding.rs
+++ b/core/src/input/binding.rs
@@ -1,0 +1,25 @@
+use crate::{Code, Modifiers};
+
+/// A keybinding used inside of a [`Keymap`](crate::Keymap).
+#[derive(Default, Copy, Clone, Debug)]
+pub struct KeyBinding {
+    /// The modifiers that have to be pressed in order to active its associated action.
+    pub modifiers: Modifiers,
+    /// The button that has to be pressed in order to active its associated action.
+    pub button: Code,
+}
+
+impl KeyBinding {
+    /// Creates a new keybinding.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use vizia_core::*;
+    /// #
+    /// let keybinding = KeyBinding::new(Modifiers::empty(), Code::KeyA);
+    /// ```
+    pub fn new(modifiers: Modifiers, button: Code) -> Self {
+        Self { modifiers, button }
+    }
+}
--- a/core/src/input/keymap.rs
+++ b/core/src/input/keymap.rs
@@ -1,10 +1,10 @@
-use crate::{Code, Context, KeyBinding, Model};
+use crate::{Code, Context, KeyChord, Model};
 use std::collections::HashMap;
 use std::hash::Hash;
 
-/// A keymap that associates keybindings with actions.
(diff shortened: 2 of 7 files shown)


rhelmot (changes requested):
My main concern here is that we're trying to cover every GUI application which could ever be built with vizia, so we want the most general possible interface here, and to me that means making keystroke <--> action a many-to-many relationship. I would also prefer if the actions didn't have to be eq + hash, so maybe the core storage should maybe be `HashMap<Key, Vec<Action>>`? For the case where you're trying to build a keymap editing view, there can be an `export` function that transforms the map into a `Vec<(Action, Vec<Key>)>`.

core/src/input/binding.rs
```diff
@@ -0,0 +1,25 @@
+use crate::{Code, Modifiers};
+
+/// A keybinding used inside of a [`Keymap`](crate::Keymap).
+#[derive(Default, Copy, Clone, Debug)]
+pub struct KeyBinding {
```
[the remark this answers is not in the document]
KDecay: Very true. I'll change this as well.

KDecay:
I renamed the `KeyBinding` to `KeyChord` and also updated the `Keymap` to allow for many-to-many relationships. It is now able to just lookup a `KeyChord` and retrieve a `Vec` over all actions that get triggered through that `KeyChord`. So instead of having to lookup every action we now just get the actions that are being pressed.

core/src/input/keymap.rs
```diff
@@ -105,131 +106,174 @@ where
...
+    /// if let Some(actions) = keymap.pressed_actions(cx, Code::KeyA) {
+    ///     for action in actions {
+    ///         println!("The action {:?} is being pressed!", action);
+    ///     }
+    /// };
     /// ```
-    pub fn remove(mut self, key: &T) -> Self {
-        self.bindings.remove(key);
-        self
+    pub fn pressed_actions(&self, cx: &Context, code: Code) -> Option<impl Iterator<Item = &T>> {
```
rhelmot, line 165: Why is this an Option<impl Iterator> and not just an iterator which is sometimes empty?

KDecay: That's because I'm returning an iterator over `&T` which in turn doesn't allow me to return something like `Vec::new().iter()`, because that value would be dropped at the end of the scope. Is there a way to get around this that you know of?

rhelmot: Can you do something like `actions.get(...).iter().flat_map(|actions| actions.iter())`? Option has an iter method which iterates over the single item if it's Some, and nothing if it's None.

KDecay: I couldn't make your suggestion work, because it would return a `FlatMap` but I found another solution is just to use [].iter() in the else case ^^ Now it just returns an `impl Iterator<Item = &T>` without an `Option` around it.

Issue #95, opened 2022-04-02
Each view's docstring should be given a once-over


rhelmot:
- [ ] Label
- [ ] Stack
- [ ] Button
- [ ] Checkbox
- [ ] Dropdown/Popup
- [ ] Element
- [ ] Image
- [ ] Knob
- [ ] List
- [ ] Menu
- [ ] NormalizedMap?
- [ ] Picker
- [ ] RadioButton
- [ ] ScrollBar
- [ ] ScrollView
- [ ] Slider
- [ ] Table
- [ ] Textbox

PR #104, opened 2022-04-04
Update label doc string

KDecay:
Part of #95.


rhelmot:
These changes are good by themselves, but can we also add a section explaining that you can pass a lens to Label::new to bind to that data?

rhelmot (approved):

PR #109, opened 2022-04-06
Add .idea to .gitignore

rhelmot:


rhelmot:
there's a local gitignore?????? this changes my life forever

robbert-vdh:
It can be configured through `git config` (check `man git-config`), but by default `$REPO_ROOT/.git/info/exclude` contains user-local ignores for a single repo, and `~/.config/git/ignore` contains ignore rules that will apply to every repo.

PR #111, opened 2022-04-07
Ensure appropriate examples build for wasm

### LINKED ISSUE #72: Ensure all appropriate examples build for wasm
Currently some of the examples don't build for wasm or don't work correctly due to missing stylesheets. All current examples should be updated to work correctly on web.

rhelmot:
Closes #72 

There is a small problem that the meaning of the `Pixels` scroll event and how it maps to your scroll wheel or trackpad actions differs across browsers. I've tuned the implementation for firefox, but there may be work to do on the winit side to smooth this over.


examples/controls/image.rs
```diff
@@ -8,6 +8,9 @@ element {
 }
 "#;
 
+#[cfg(target_arch = "wasm32")]
+compile_error!("This example uses image loading that does not work on the web");
```
geom3trik, line 12: I looked into this a bit before when trying to make a web app with images. The part that doesn't work on web is actually spawning a thread, the image loading itself works. However, I also discovered that the image only works when _not_ set to repeat in x and y.

Oh and loading an image from the web doesn't work either, but I think it could with an async version.

winit/src/application.rs
```diff
@@ -101,12 +102,6 @@ impl Application {
         self
     }
 
-    pub fn locale(mut self, id: &str) -> Self {
```
geom3trik, line 104: This is something we can revisit later but I think it would be good to be able to set the locale for testing purposes.

geom3trik (approved):
I'm happy for this to be merged now and we can come back to the image example if we want to get something working for wasm.

PR #112, opened 2022-04-07
really export femtovg contents instead of whole module

rhelmot:


geom3trik:
Ah oops, do we need to do the same for image too?

geom3trik (approved):

PR #119, opened 2022-04-09
Update README.md

geom3trik:


README.md
```diff
@@ -1,148 +1,109 @@
...
-	count: i32,
-}
 
-impl Model for AppData {}
+## Features
+ - Multiplatform (Windows, Linux, MacOS, Web)
+ - Declarative API
+ - Reactive with an elm-like archtecture
+ - Flexible layout engine, powered by [morphorm](https://github.com/vizia/morphorm)
+ - CSS-like styling with hot reloading
```
rhelmot, line 75: We're not css-like, we're actually using css! I don't think that just because we don't have the same properties as html we have to downplay it here.

README.md
```diff
@@ -1,148 +1,109 @@
...
+ - Style property animations
 
-```
-The `Lens` derive macro enables binding views to fields of some state.
+## Learning Vizia
+<!-- Add docs to this when ready -->
+<div align="center">
+  <h3>
+    <a href="https://book.vizia.dev/"> Getting Started </a>
```
rhelmot, line 82: Do you want to link to something else for Getting Started? Personally, I would remove the link entirely.

README.md
```diff
@@ -1,148 +1,109 @@
...
-		if let Some(app_event) = event.message.downcast() {
-			match app_event {
-				AppEvent::Increment => self.count += 1,	
-				AppEvent::Decrement => self.count -= 1,
-			}
-		}
-	}
-}
-```
+Example projects go here... with pictures and stuff...
```
rhelmot, line 93: Do you want me to make arborio look pretty so you can put it here?

README.md
```diff
@@ -1,148 +1,109 @@
...
-			Label::new(cx, "Decrement")
-		});
-
-		Binding::new(cx, AppData::count, |cx, some_data|{
-			Label::new(cx, &count.get(cx).to_string());
-		});
-	});
-}).run();
-```
+Vizia is also being used for the [Meadowlark project](https://github.com/MeadowlarkDAW/Meadowlark). An attempt to build a modern Digital Audio Workstation (DAW) with Rust.
```
rhelmot, line 95: Comma instead of period here

README.md
```diff
@@ -1,148 +1,109 @@
...
+
+## Architecture
```
rhelmot, line 97: Suggested text:

```
Vizia's core abstraction is a tree of **entities**. Each entity is just an integer, but this integer keys a number of dictionaries affecting how the entities look and act:
- **Style properties**, directives for layout and drawing
- **Models**, custom user-stored data which store a list of observers
- **Views**, bundles of callback logic to handle custom drawing, data reactivity, and user interaction

In order to interact with this tree of entities, there are several tools. The important ones are:
- **Bindings**, special views which watch for data changes and refresh their children
- **Events**, messages that propagate up or down the tree and trigger callbacks

To learn more, read the book, check out the examples, or join our discord!
```

README.md
```diff
@@ -1,148 +1,109 @@
...
+
+## Roadmap
```
rhelmot, line 99: I think this should just be a list of tracking issues

README.md
```diff
@@ -1,148 +1,109 @@
...
+
+## License and Attribution
+Vizia is licensed under [MIT](https://github.com/vizia/vizia/blob/main/LICENSE).
+
+Fonts used in Vizia:
+- Roboto licensed under [Apache license](http://www.apache.org/licenses/LICENSE-2.0).
+- Entypo licensed under CC BY-SA 4.0.
+- Material Icons licensed under [Apache license](https://github.com/google/material-design-icons/blob/master/LICENSE).
```
rhelmot, line 107: what about the emoji and arabic fonts?

README.md
rhelmot: If you want to make it clear that we don't support the same style properties as html I would say "HTML-like CSS styling"

geom3trik: Okay good point. Though it's not just the difference in properties, there's quite a few selector things we don't support (yet) as well.

README.md
geom3trik: My thinking was to link to the first section of the book which you've written which has the really nice set-up stuff. But yeh I guess this would just be a duplicate of the "The Book" link. I'm just thinking that new users might be more likely to look for a "Getting Started" link than a book link.

README.md
geom3trik: I would definitely love to put arborio here but there's no rush if styling is something you're leaving till later.

README.md
geom3trik: I like this very much, thank you!

geom3trik: I've decided to remove the architecture section and will instead dedicate a whole file to it (eventually).

rhelmot (approved):

PR #123, opened 2022-04-14
Multiwindow

geom3trik:
Adds the basics for multiwindow.

TODO:

- Handle window events correctly for each window
- Apply styling/layout/rendering to each window only when required
- Apply hover for each window separately


geom3trik:
Superseded by #438

PR #130, opened 2022-04-19
Make localization more expressive

rhelmot:
- Lets you specify constant values as fluent args
- Makes all keys passed as String instead of &'static str. maybe a Cow would be more appropriate? idk
- Re-exports FluentValue so users don't have to depend on fluent-bundle to make types compatible with us


geom3trik:
I'm wondering also if maybe there should be an option to default to the key if the translation is missing? Probably that shouldn't be the default but it could be useful to have it as an option.

geom3trik (approved):

Issue #129, opened 2022-04-19
Add a "null" backend


rhelmot:
We should have the ability to construct a backend without winit or baseview installed. There should still be a canvas, but writes to it should be discarded, or if that's not an option, written to an offscreen buffer. This will allow us to build testcases (including doc tests and perf tests) that can run in CI.

PR #135, opened 2022-04-21
Refactor convert methods in `winit` crate and small changes

KDecay:
I refactored the convert methods in the `winit` crate and put them all into one module that's called `convert` instead of having multiple modules for converting things. I also made some smaller changes like putting `use`s together and changing `no_run` examples to be run so we can assure that the examples work.


rhelmot:
I'm curious to see what CI says about this one :) I think if you try running examples that actually build an application in CI, it'll crash, since there is no windowing system to render to. See #129

KDecay:
@rhelmot Oh interesting. I didn't think about that. I removed the `run()` call because that would cause an infinite loop, but if even the construction already crashes I'll just add the `no_run` again. I somehow mixed `ignore` and `no_run` and thought `no_run` also doesn't check if the code compiles.

rhelmot:
oh, I didn't know about the difference between no_run and ignore. but yes, it seems that it does panic: https://github.com/vizia/vizia/runs/6116533461?check_suite_focus=true#step:8:541

KDecay:
Yeah. I'll add the `no_run` flag again :D

rhelmot (approved):

PR #138, opened 2022-04-21
knobs: make arc_start and arc_stop settable with ArcTrack::new()

Fredemus:
Doesn't change anything when using `Knob::new()` just makes it possible to "rotate" an `ArcTrack` when using `Knob::custom()`


rhelmot (approved):

geom3trik:
I think this is fine for now but in the future we might want to think about turning these into modifiers on the view rather than inputs.

PR #139, opened 2022-04-22
Update `Checkbox` doc string

KDecay:
I updated the doc string of the `Checkbox` and also ported some things from the [book](https://github.com/vizia/vizia-book/blob/develop/src/views/checkbox.md). I wasn't sure what the `getting` example should showcase or what it would be useful for so I didn't add it for now. If it is something useful then please let me know so I can add it.

### DIFF
--- a/core/src/views/checkbox.rs
+++ b/core/src/views/checkbox.rs
@@ -2,41 +2,114 @@ use crate::{Context, Handle, Lens, MouseButton, Res, View, WindowEvent};
 
 pub const ICON_CHECK: &str = "\u{2713}";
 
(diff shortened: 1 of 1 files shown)


rhelmot (approved):

geom3trik:
What's the `getting` example?

KDecay:
@geom3trik Inside of the book linked in the PR there is an example for "`get`ting the lens".

geom3trik:
Oh I see. Yeh, better to keep things simple. We don't need to `get` the lens just to demo the checkbox callback.

PR #142, opened 2022-04-23
Replace `WindowDescription` argument of `Application` with modifiers

geom3trik:
Before:
```rust
Application::new(WindowDescription::new().with_title("Title"), |cx|{
    ...
}).run();
```

After:
```rust
Application::new(|cx|{
    ...
})
.title("Title")
.run();
```


winit/src/window.rs
```diff
@@ -155,27 +155,85 @@ impl View for Window {
...
+            WindowEvent::SetTitle(title) => {
+                self.window().set_title(title);
+            }
+
+            WindowEvent::SetSize(size) => {
+                self.window().set_inner_size(LogicalSize::new(size.width, size.height));
+            }
+
+            WindowEvent::SetMinSize(size) => {
+                self.window().set_min_inner_size(Some(LogicalSize::new(size.width, size.height)));
```
KDecay, line 167: You are not able to remove the `min_inner_size` currently right? I think it would make more sense to pass an `Option` here instead.

winit/src/window.rs
```diff
@@ -155,27 +155,85 @@ impl View for Window {
...
+            }
+
+            WindowEvent::SetMaxSize(size) => {
+                self.window().set_max_inner_size(Some(LogicalSize::new(size.width, size.height)));
```
KDecay, line 171: Same here

core/src/views/textbox.rs
```diff
@@ -344,12 +344,13 @@ impl Model for TextboxData {
             }
 
             TextEvent::Submit => {
-                cx.emit(TextEvent::EndEdit);
+                println!("{}", self.text);
```
KDecay, line 347: This should probably be removed or is there a good reason for it to be there?

KDecay (commented):
I definitely love this change. Looks a lot cleaner.

winit/src/window.rs
geom3trik: Ah yes, good point!

core/src/views/textbox.rs
geom3trik: Yep my bad, needed to debug something.

KDecay (approved):

PR #143, opened 2022-04-24
Create a `backends` folder for `winit` and `baseview`

KDecay:
I moved the `winit` and `baseview` crates into a `backends` folder to make the intent clearer and also group our backends nicely. This may also come in handy if we decide to add more or different backends to vizia.


KDecay:
Closing this out in favor of #171.

PR #146, opened 2022-04-25
Update `dropdown` doc string

KDecay:
I updated the `dropdown` doc string with the example from the [book](https://github.com/vizia/vizia-book/blob/develop/src/views/dropdown.md).

### DIFF
--- a/core/src/views/dropdown.rs
+++ b/core/src/views/dropdown.rs
@@ -46,8 +46,84 @@ use crate::views::popup::PopupData;
 /// .width(Pixels(100.0));
 /// ```
 ///
-/// The line marked "close the popop" is not required for anything other than closing the popup -
+/// The line marked "close the popup" is not required for anything other than closing the popup -
 /// if you leave it out, the popup will simply not close until the user clicks out of the dropdown.
+///
+/// ## Custom Dropdown
+///
+/// The dropdown doesn't have to be the current state and then a set of options - it can contain any
+/// set of views in either location. Here's an example where you can use a textbox to filter a list
+/// of checkboxes which pop up when you click the textbox:
+///
+/// ```
+/// # use vizia_core::prelude::*;
+/// # let cx = &mut Context::new();
+/// #
+/// # cx.add_theme(r#"
+/// #     dropdown popup {
+/// #         background-color: white;
+/// #     }
+/// # "#);
+///
+/// #[derive(Lens, Clone, PartialEq, Eq)]
+/// struct AppData {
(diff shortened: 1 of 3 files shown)


core/src/views/dropdown.rs
```diff
@@ -3,11 +3,67 @@ use crate::{
     View,
 };
 
+/// The dropdown icon of a dropdown view.
 pub const ICON_DOWN_OPEN: &str = "\u{e75c}";
 
-pub struct Dropdown {}
+/// A dropdown used to display some state with the ability to open a popup with options to change that state.
```
geom3trik, line 9: Missing 'is'. "A dropdown **is** used to display..."

core/src/views/dropdown.rs
```diff
@@ -3,11 +3,67 @@ use crate::{
...
+///
+/// Usually a dropdown is used in the context of a "combo box" or "list picker" to allow the user to select
```
geom3trik, line 11: I could be wrong but I always thought a combo box was specifically a dropdown + textbox combination.

core/src/views/dropdown.rs
```diff
@@ -3,11 +3,67 @@ use crate::{
...
+/// #
+/// Dropdown::new(
+///     cx,
+///     |cx| Label::new(cx, AppData::value),
+///     |cx| {
+///         for i in 0..5 {
+///             Label::new(cx, i)
+///                 .on_press(move |cx| {
+///                     cx.emit(AppEvent::SetValue(i));
+///                     cx.emit(PopupEvent::Close); // close the popup
```
geom3trik, line 45: Could be worth mentioning here that closing the popup when selecting is optional. This allows for something like a dropdown where multiple things can be selected.

geom3trik (changes requested):
I think we should add some more detail about what can be done with the dropdown. 
The first closure doesn't have to show the current option, it could just display some text, or an icon, or text + icon, or even a textbox.
The second closure is for the contents but it doesn't necessarily have to be a list or even a collection of the same views. For example, if you wanted a dropdown with multiple selectable items where each one has a checkbox and a label then you could build that.
The dropdown then is the bare bones of a view which has a "closed" view and an "open" view.

core/src/views/dropdown.rs
rhelmot: I'm pretty sure you could use the Dropdown widget to implement that kind of form element as well!

rhelmot (approved):

geom3trik (approved):

PR #147, opened 2022-04-25
Add icon to window modifiers

geom3trik:


rhelmot:
Do we maybe want to make this take a femtovg::ImageSource?

geom3trik:
I'm not sure it's necessary in this case because  an image can be loaded with the image crate and then passed directly to winit. Femtovg doesn't need to be involved. Is there an advantage to using `femtovg::ImageSource` here?

rhelmot:
I was thinking just for api consistency, reusing an exported interface, and for the fact that it's not trivial usage-wise to get from e.g. a .png to an input buffer

geom3trik:
With how it is currently you can do this:
```rust
 let​ icon ​=​ image​::​open​(​"path_to_image.png"​).​unwrap​();

Application::new(|cx|{}).icon(icon.into_bytes(), icon.width(), icon.height());
```

What would it be like using `ImageSource`?

rhelmot:
`Application::new(|cx|{}).icon(image::open("path_to_image.png).unwrap().into())`

I suppose that's not a huge improvement in usability, especially since we're re-exporting the image crate. I just have a personal distaste for requiring the user to handle buffers of image data like that, but I'm willing to put that aside.

rhelmot (approved):
As discussed on discord, we probably won't be able to do the thing I wanted :)

PR #149, opened 2022-04-26
Update `Element` doc string

KDecay:
I updated the `Element` doc string with some examples I could come up with. I also saw that you are able to bind an `Element` to some data, but I'm not quite sure how exactly that works.

### DIFF
--- a/core/src/views/element.rs
+++ b/core/src/views/element.rs
@@ -13,7 +13,7 @@ use crate::prelude::*;
 /// create an invisible spacer between other views.
 ///
 /// ```
-/// # use vizia_core::*;
+/// # use vizia_core::prelude::*;
 /// #
 /// # let cx = &mut Context::new();
 /// #
@@ -27,7 +27,7 @@ use crate::prelude::*;
 /// An element can be used to display a rectangle like this black one with a size of 100 by 100 pixels.
 ///
 /// ```
-/// # use vizia_core::*;
+/// # use vizia_core::prelude::*;
 /// #
 /// # let cx = &mut Context::new();
 /// #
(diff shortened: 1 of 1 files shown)


core/src/views/element.rs
```diff
@@ -2,17 +2,91 @@ use crate::{Context, Handle, View};
 
 /// A basic element with no interactivity.
 ///
+/// An element has no children and no special rendering logic. It can be used to render a rectangle,
+/// a line or a circle using view styling rules.
```
geom3trik, line 6: Another useful thing you can do with an element is add a image backgound or a gradient.

core/src/views/element.rs
```diff
@@ -2,17 +2,91 @@ use crate::{Context, Handle, View};
...
 ///
-pub struct Element {}
+/// # Examples
+///
+/// ## Element without styling
+///
+/// An element can be used without styling which displays nothing at all.
```
geom3trik, line 12: While this is true I don't know what the use would be.

geom3trik, line 12: You could use an element as an invisible spacer between other views.

core/src/views/element.rs
```diff
@@ -2,17 +2,91 @@ use crate::{Context, Handle, View};
...
+///
+/// ```
+/// # use vizia_core::*;
+/// #
+/// # let cx = &mut Context::new();
+/// #
+/// Element::new(cx)
+///     .width(Pixels(100.0))
+///     .height(Pixels(100.0))
+///     .border_radius(Pixels(50.0))
```
geom3trik, line 66: A better way to do this would be `Percentage(50.0)` for the border radius.

geom3trik (changes requested):

geom3trik (approved):

PR #154, opened 2022-05-15
Allow custom kinds of bindables and stores

rhelmot:


core/src/state/binding.rs
```diff
@@ -1,141 +1,264 @@
...
-                        }
-                    } else {
-                        let mut observers = HashSet::new();
-                        observers.insert(id);
+        self.view(cx, |v| v.cloned())
+    }
 
-                        let model = model_data.downcast_ref::<L::Source>().unwrap();
+    fn requests(&self) -> Vec<StoreId>;
+    fn make_store(&self, source: ModelOrView) -> Option<Box<dyn StoreHandler>>;
```
geom3trik, line 127: Could it be worth using an associated type for the returned store?

core/src/state/binding.rs
```diff
@@ -1,141 +1,264 @@
...
+                } else {
+                    let idx = model_data_store.lenses_dup.len();
+                    model_data_store.lenses_dup.push(bindable.make_store(model_data).expect(
+                        "Bug: bindable failed to produce a store. Probably an internal error.",
+                    ));
+                    model_data_store.lenses_dup.get_mut(idx).unwrap()
+                };
+
+                // ...so that we can register the observer with the appropriate store.
+                bindable.add_to_store(store.as_mut(), id);
```
geom3trik, line 97: Why is this a method on `bindable` rather than on `store`? Would it not make more sense to have `store.add_observer(id)`?

core/src/state/binding.rs
```diff
@@ -1,141 +1,264 @@
...
+        )
+    }
 
-                        if let Some(key) = state.lens.cache_key() {
-                            model_data_store.lenses_dedup.insert(key, state);
-                        } else {
-                            model_data_store.lenses_dup.push(state);
-                        }
-                    }
+    fn requests(&self) -> Vec<StoreId> {
```
geom3trik, line 152: I'm a bit confused how this works because the `make_store` method seems to only be able to create a single store but this method suggests that a bindable can link to many stores.

core/src/state/binding.rs
```diff
@@ -1,141 +1,264 @@
...
+    fn add_to_store(&self, store: &mut dyn StoreHandler, entity: Entity) {
+        self.child.add_to_store(store, entity);
     }
 }
 
-impl<L: 'static + Lens> View for Binding<L> {
-    fn element(&self) -> Option<&'static str> {
-        Some("binding")
+#[derive(Clone, Debug)]
+pub struct ConstantBindable<T> {
```
geom3trik, line 232: What is this for?

core/src/views/scrollview.rs
```diff
@@ -106,12 +105,12 @@ impl<L: Lens<Target = ScrollData>> ScrollView<L> {
     {
         VStack::new(cx, content)
             .class("scroll_content")
-            .bind(data.clone(), |handle, data| {
-                let data = data.get(handle.cx);
-                let left = ((data.child_x - data.parent_x) * data.scroll_x).round();
-                let top = ((data.child_y - data.parent_y) * data.scroll_y).round();
-                handle.left(Units::Pixels(-left)).top(Units::Pixels(-top));
-            })
+            .left(data.clone().map_shallow(|data| {
```
geom3trik, line 108: So does this mean that `map_shallow` replaces the need for `.bind()` when using it to set multiple properties?

core/src/state/binding.rs
rhelmot: I considered this, but I'd like to keep the flexibility of a single bindable being able to bind to multiple kinds of stores!

core/src/state/binding.rs
rhelmot: make_store will be called once for each request!

core/src/state/binding.rs
GMAtkinson: But how would a single bindable choose what store to use?

rhelmot: The thing I was envisioning is that there would be a MultiBindable which contains two child bindables. It will then dispatch its method calls to BOTH of the children, and take the first result that succeeds.

core/src/state/binding.rs
GMAtkinson: Ah I see! But is the observer not always an entity?

core/src/state/binding.rs
GMAtkinson: Ah okay, so the requests vector defines what stores the binable links to and then the `make_store` functions creates the store for each item?

core/src/state/binding.rs
rhelmot: The observer will always be an entity, but the observer's metadata (e.g. the list index being bound to) can be of arbitrary type.

core/src/state/binding.rs
rhelmot: Correct

core/src/state/binding.rs
GMAtkinson: Oh okay, that's an interesting way to do it. I was thinking that this would be done with just the `MultiBindable` type which would have an associated store that holds on to two lenses and two pieces of state but one list of observers. Then the output of the `Bindable` would be a tuple of the output types for the lenses.

rhelmot: I think that approach would require that the two lenses both be on models which are built on the same entity, which is not desirable

PR #156, opened 2022-05-17
Fix for keyboard modifiers and cursor icon when releasing capture

geom3trik:
This also fixes copy/paste for the textbox as well as the cursor icon not changing when clicking out of the textbox.


rhelmot (dismissed):

rhelmot (approved):
This looks good. the only change I would maybe make is to attach the locking to the current entity so the lock automatically goes away if the entity is removed.

PR #157, opened 2022-05-20
Fix (most) Clippy Warnings

geom3trik:


rhelmot (approved):
There's no way I can look through all of this but if it's a strict improvement to what clippy says it seems good to me.

PR #160, opened 2022-06-14
Style inheritance bugfix

geom3trik:


geom3trik:
This was added in #237

PR #161, opened 2022-06-15
Fix assorted issues

geom3trik:
- Expose `Key`
- Add `set_hover` method to context
- Add bool to textbox `on_submit` to distinguish between enter and defocus
- Scale border by scale factor for layout

### DIFF
--- a/core/src/views/scrollview.rs
+++ b/core/src/views/scrollview.rs
@@ -107,10 +107,15 @@ impl<L: Lens<Target = ScrollData>> ScrollView<L> {
         VStack::new(cx, content)
             .class("scroll_content")
             .bind(data.clone(), |handle, data| {
-                let data = data.get(handle.cx);
-                let left = ((data.child_x - data.parent_x) * data.scroll_x).round();
-                let top = ((data.child_y - data.parent_y) * data.scroll_y).round();
-                handle.left(Units::Pixels(-left)).top(Units::Pixels(-top));
+                let dpi_factor = handle.cx.style().dpi_factor;
+                if dpi_factor > 0.0 {
+                    let data = data.get(handle.cx);
+                    let mut left = ((data.child_x - data.parent_x) * data.scroll_x).round()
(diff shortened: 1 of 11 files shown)


robbert-vdh:
Forgot to post here, but this together with #160 fixes all of the issues I had with with the vizia GUIs in NIH-plug.

core/src/views/scrollview.rs
```diff
@@ -107,10 +107,15 @@ impl<L: Lens<Target = ScrollData>> ScrollView<L> {
...
-                handle.left(Units::Pixels(-left)).top(Units::Pixels(-top));
+                let dpi_factor = handle.cx.style().dpi_factor;
+                if dpi_factor > 0.0 {
+                    let data = data.get(handle.cx);
+                    let left = ((data.child_x - data.parent_x) * data.scroll_x).round()
+                        / handle.cx.style().dpi_factor as f32;
+                    let top = ((data.child_y - data.parent_y) * data.scroll_y).round()
+                        / handle.cx.style().dpi_factor as f32;
+                    handle.left(Units::Pixels(-left.abs())).top(Units::Pixels(-top.abs()));
+                }
```
rhelmot, line 118: What does this fix?

core/src/views/textbox.rs
```diff
@@ -336,18 +336,23 @@ impl Model for TextboxData {
...
             }
 
             TextEvent::EndEdit => {
                 self.edit = false;
+                cx.set_checked(false);
+                cx.release();
             }
 
-            TextEvent::Submit => {
+            TextEvent::Submit(flag) => {
```
rhelmot, line 353: Can we pick a more descriptive name than "flag"?

core/src/views/scrollview.rs
geom3trik: The cache contains values in physical coordinates but the style properties expect logical. This fixes an bug where extra space gets put at the bottom of a scrollable area.

rhelmot (approved):

PR #164, opened 2022-06-18
Use operator overloading for lens entension methods

geom3trik:
Allow this:
```rust
AppData::subdata.then(Subdata::field)
```
To be expressed like this:
```rust
AppData::subdata >> Subdata::field
```


rhelmot:
I'm strongly opposed to operator overloading for this purpose. I think it's one of the least intuitive parts of "basic" C++ and would dramatically reduce learnability. `then` is clean enough as is.

geom3trik:
The more I think about it the more I would agree with that. It was a fun idea but it's definitely just less intuitive.

PR #167, opened 2022-06-22
Document `mouse.rs` and add `frame_delta` method

KDecay:
I documented the `mouse.rs` file and also added a `frame_delta` method that returns the delta of the mouse cursor position of the current and previous frame. This is useful if you want to use this as a value that gets added to some other value every frame. 

An example would be dragging the timeline in a `DAW` like `Meadowlark`. The `delta` method wouldn't work in this case, because you would either have to store the previous mouse cursor position or the value you want to increase/decrease which is rather inconvenient.

I also added an example for the usage of both the `frame_delta` and `delta` methods.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,6 +183,10 @@ name = "keymap_change_entries"
 path = "examples/keymap_change_entries.rs"
 
 [[example]]
+name = "mouse_delta"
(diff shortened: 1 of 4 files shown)


rhelmot (approved):

rhelmot:
Looks great. Thanks for all your work!

PR #174, opened 2022-07-03
Remove Rc from Map lens

geom3trik:


rhelmot:
Can the user still have the option somehow to use a rc if they want to? Is there some sort of "callable rc" type?

GMAtkinson:
What would be the use case? Is there an advantage to using an `Rc`?

geom3trik:
What would be the usecase? Is there an advantage to using Rc?

rhelmot:
Yes, you don't have to implement Clone for the closure's contents. I'm not sure if this is a use case which is worth designing for, I just want to make sure I know what kind of restrictions we're putting on the interface.

rhelmot (approved):

rhelmot:
I decided that comment was me being silly. If there's ever any case you want to put non-clone data in a lens with an rc, you can just put that specific data into an rc. That's better on pretty much all levels.

PR #178, opened 2022-07-19
Add `EventContext`

geom3trik:
~~!!!DO NOT MERGE!!! - Temporarily disabled image loading~~


rhelmot (approved):
Looks good so far. None of the interface changes seem controversial.

rhelmot (commented):
Looking good!

rhelmot (approved):
let's goooooooooooooo

I'm okay with merging this now. You mentioned on discord you wanted to make members on Context crate-private, will this also necessitate having a BuildContext?

geom3trik:
> You mentioned on discord you wanted to make members on Context crate-private, will this also necessitate having a BuildContext?

In my mind no. There will still be methods on `Context` to do things like sending events, but if the user wants to do something that requires access to things on `EventContext` then they can use `on_build` modifier which will give access to the event context. 

But I'll save making anything on `Context` crate private for another PR.

rhelmot (approved):

PR #180, opened 2022-07-20
Cleanup action modifiers and add docs

geom3trik:


core/src/modifiers/actions.rs
```diff
@@ -506,130 +423,91 @@ impl<V: View> View for Geo<V> {
...
 
-    fn on_hover<F>(self, action: F) -> Handle<'a, Hover<Self::View>>
+    /// Adds a callback which is performed when the left mouse button is released on a view after being pressed.
+    fn on_release<F>(self, action: F) -> Handle<'a, Release<V>>
     where
         F: 'static + Fn(&mut Context);
 
-    fn on_over<F>(self, action: F) -> Handle<'a, Over<Self::View>>
+    /// Adds a callback which is performed when the mouse pointer moves over or away from a view.
+    fn on_hover<F>(self, action: F) -> Handle<'a, Hover<V>>
```
rhelmot, line 439: I'm generally not in favor of multiplexing interfaces like this. How about on_hover_in, on_hover_out, etc?

rhelmot: If there's precedent for this kind of interface in other frameworks I could be swayed.

geom3trik: The inspiration is from siftui: <https://developer.apple.com/documentation/swiftui/view/onhover(perform:)>. But I think I agree that this isn't necessarily better, since it now requires an `if` statement every time you want to do something on hover. I'd like to keep the names as simple as possible, so how about `on_hover` vs `on_leave`?

rhelmot: The problem with on_hover and on_leave is it's more things to memorize or look up. It would be okay if it was hover/leave, but we would need to come up with good names for hover in/out, over in/out, and focus in/out.

geom3trik: You make a good point. I think one reason to maybe keep it as `on_hover`, `on_over` and `on_focus` is to match the corresponding css pseudoclass selectors. But then if we have `on_hover_out`, `on_focus_out` and `on_over_out` would this be more confusing because of the asymmetry?

rhelmot: I think the confusingness is mitigated by the fact that when you do a docs search for the css selector, the right things come up :)

geom3trik:
Okay , I've gone and split the action modifiers into `on_hover`/`on_hover_out`, `on_over`/`on_over_out`, and `on_focus`/`on_focus_out`. I'm still not sure about the names though.

While working on this two other questions came to mind:
 1. Should these actions be part of the context? So instead of removing the view and then replacing it with something like `Hover<V>`, we instead have something like `hover_actions: HashMap<Entity, Box<dyn Fn(&mut Context)>>` somewhere in the context.
 2. Should these modifiers actually be part of an `ActionModifiers` trait (as it is now) or should they just be methods on `Handle`? The advantage of the former could be better docs but then I *think* you can achieve the same with a separate `impl Handle` block within its own module.

rhelmot:
1. I think, and don't quote me on this, you could get away with having e.g. a `HoverActions` model? I would prefer that. Regardless, I would definitely prefer pretty much anything to the `Hover<V>` approach :sweat_smile: 
2. It would probably be cleaner to not have a separate trait, I don't know whether there's any sort of community consensus on which pattern people prefer using.

geom3trik:
Changed action modifiers to use a model now instead of wrapping views. So adding an action modifier like `on_press()` will add the callback to a model attached to the view and the model calls the callback when it receives the relevant window event.

A couple of questions to answer:

1.  Should each action modifier have its own model instead of them all sharing one model?
2. Should action modifiers and window event names be synchronised. Some of them are, but something like `on_hover` action uses the `MouseEnter` window event?

rhelmot:
1. I believe they should all share one model to reduce contention in the model type id -> model hash table
2. They should definitely be synchronized.

geom3trik:
> 1. I believe they should all share one model to reduce contention in the model type id -> model hash table

One potential issue with having the one model is that each action is a function pointer. So adding just one action will add a pointers worth of memory for each action in the model. Imagine we have a long list of views, each with one action, then that's potentially a lot of wasted memory. Though I admit the list would need to be very large for it to make a dent on a modern computer, and we could solve this with virtualized lists.

geom3trik:
With #246 merged I've now updated the action modifiers to sync better with the window events. The last one to potentially change is the `on_hover` and `on_hover_out` actions which don't currently match the window event. However, the reason for this is because they instead match the pseudoselector for styling. We could rename the window events to match but in this case I'm tempted to leave it as it is. Interested to hear any thoughts on this.

crates/vizia_core/src/modifiers/actions.rs
```diff
@@ -1,634 +1,551 @@
...
-        event.map(|window_event, _| match window_event {
-            WindowEvent::Press { mouse } => {
-                let over = if *mouse { cx.hovered() } else { cx.focused() };
-                if cx.current() != over && !over.is_descendant_of(cx.tree, cx.current()) {
-                    return;
-                }
-                if let Some(action) = &self.action {
-                    (action)(cx);
-                }
+        event.map(|actions_event, _| match actions_event {
```
rhelmot, line 41: this seems like the perfect opportunity for `event.take()`. Would that also let us get rid of the `Arc`?

geom3trik: Good idea! I think I originally wrote this part before `event.take()` was merged and then I completely forgot it existed. I'll give this a go and see if we can renove those `Arc`s.

geom3trik: I made the suggested change to use `event.take()` and I was able to remove the clones but I still need to `Box` the closures unless I'm missing something? In any case it did remove the need for `Arc`.

geom3trik:
Supersedes #245 and fixes #232

rhelmot (approved):

PR #182, opened 2022-07-28
Update checkbox style and example

geom3trik:
- Update checkbox styling.
- Update the checkbox example to show off new styling.
- Fix a bug where multiple pseudoclass selectors weren't matching correctly. For example:

This rule...
```css
checkbox:checked:hover {
    background-color: red;
}
```
should have a higher specificity than this rule...
```css
checkbox:hover {
    background-color: green;
}
```
but previously the former rule wasn't matching at all on the element, causing the latter rule to be selected in error. This is now fixed.

### DIFF
--- a/core/resources/themes/default_theme.css
+++ b/core/resources/themes/default_theme.css
@@ -62,11 +62,9 @@ button.accent:active label {
 checkbox {
     width: 20px;
     height: 20px;
-    border-radius: 3px;
+    border-radius: 1px;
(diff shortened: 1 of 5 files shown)


geom3trik:
This PR also changes the cursor icon on hover for disabled views. The cursor icon is now not set for diabled views but I'm not sure whether or not this is the 'correct' approach.

rhelmot (approved):

Issue #55, opened 2022-03-02
Focus order keyboard navigation


geom3trik:

rhelmot:
Things necessary to close this:

- visual indicators for which element is focused. I think we should add the `outline` property from html, add a style for it on `*:focus`, and override it on any element which has custom focus.
- elements need to respond to spacebar like a click. I think we should add a `handle.press_or_space(action)`.

geom3trik:
Good idea for the `outline` style property, I'll go ahead and add that.

I think the key used to trigger an action should be customizable and is dependent on the control. Not sure if this should be specific to the control itself or as an `Enviroment` variable which can be configured for all controls.

0HyperCube:
Other things necessary for good keyboard navigation:
- [ ] Popups (such as the modal example) should lock focus to within them so you can't tab outside until they are closed.
- [ ] Dropdowns should allow selection via the up and down arrow keys and enter/space like html `<select>` - this requires changing how dropdowns are done because we currently have no concept of a selected row.
- [ ] Menus should be also be navagatable with the arrow keys with left and right opening or closing submenus or navigating to the next or previous menu stack.
- [ ] Arrow key support for changing knobs
- [ ] Arrow key support for changing sliders
- [ ] A general way of making stacks or lists navigatable via arrow keys.

geom3trik:
This is a very useful list of requirements, thank you. I think if we sort out keyboard navigation within lists then that can be used for both menus and dropdowns.

PR #185, opened 2022-08-01
Focus improvements

0HyperCube:
- Focus style updates displays immediately
- The `focus_order` example always keeps your focus on the buttons
- Add `KEYBOARD_NAVIGATABLE` ability to allow for focusable elements that are not in the tab order (the equivalent of `tabindex="-1"` in html)
- Buttons get focused on click (this triggers the focus pseudo class which equivalent behaviour to html - html also has the `:focus-visible` pseudo class to only show the style when using keyboard navigation which we may also want)

### DIFF
--- a/core/src/context/mod.rs
+++ b/core/src/context/mod.rs
@@ -1115,7 +1115,7 @@ impl Context {
                             prev_focused
                         } else {
                             TreeIterator::full(&self.tree)
-                                .filter(|node| is_focusable(&self.style, *node))
+                                .filter(|node| is_navigatable(&self.style, *node))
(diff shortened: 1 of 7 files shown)


geom3trik (changes requested):
This looks great! Thank you for your contribution. Just a couple of very minor changes and then this is good to be merged. I agree that a `:focus-visible` pseudo-selector should be added next, along with an `outline` property, which is on my todo list.

geom3trik (approved):

geom3trik:
Related to #55

PR #189, opened 2022-08-03
Add backend context trait and cleanup context

geom3trik:


rhelmot (commented):
This all looks fine, except that I'm not really sure what the point of BackendContext is? It's not hiding any implementation details, it's just moving interfaces from Context onto a trait, right?

geom3trik:
> This all looks fine, except that I'm not really sure what the point of BackendContext is? It's not hiding any implementation details, it's just moving interfaces from Context onto a trait, right?

It doesn't hide implementation details but it does hide some methods from users that are only needed by winit/baseview. Note that `BackendContext` is not part of the prelude so a user importing the prelude won't be able to access these methods on context. I thought about making backend context a struct to ensure this but unfortunately the backend (winit/baseview) needs to make the context and pass it to the application closure, so to make the members of context crate private it needs to be a trait.

So basically it cleans up context and should make the docs look a bit nicer.

rhelmot:
In that case, I'm going to suggest that we move these methods from a trait to just plain functions which take a context as an argument. This is because some IDEs (mine is clion) will autocomplete methods from non-imported public traits.

geom3trik:
> In that case, I'm going to suggest that we move these methods from a trait to just plain functions which take a context as an argument. This is because some IDEs (mine is clion) will autocomplete methods from non-imported public traits.

Okay, I've made backend context a struct now so hopefully this works better.

rhelmot (approved):

PR #191, opened 2022-08-05
Do not rebuild entities if their parent has been rebuilt

rhelmot:
This fixes an incredibly annoying and incredibly difficult to pin down bug I was lucky enough to reproduce in arborio this morning. Basically, if you have a view which is only ever used in the context of a parent view and whose builder assumes the conditions necessary to build the parent are satisfied, and the parent gets removed because its preconditions are invalidated AND the child gets marked for rebuild in the same cycle, the child may panic while rebuilding because it should not be being rebuilt!

The fix is to a) rebuild observers in tree order and b) keep track of which entities have been removed, and if so, skip them during the observer update process.

### DIFF
--- a/core/src/context/mod.rs
+++ b/core/src/context/mod.rs
@@ -88,6 +88,8 @@ pub struct Context {
     click_pos: (f32, f32),
 
     pub ignore_default_theme: bool,
+
+    removed_entities: HashSet<Entity>,
 }
 
 impl Context {
@@ -131,6 +133,7 @@ impl Context {
             click_pos: (0.0, 0.0),
 
             ignore_default_theme: false,
+            removed_entities: HashSet::new(),
         };
 
         Environment::new().build(&mut result);
@@ -461,6 +464,8 @@ impl Context {
                 self.captured = Entity::null();
             }
         }
+
+        self.removed_entities.extend(delete_list);
     }
 
     /// Send an event containing a message up the tree from the current entity.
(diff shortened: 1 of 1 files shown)


core/src/context/mod.rs
```diff
@@ -88,6 +88,8 @@ pub struct Context {
     click_pos: (f32, f32),
 
     pub ignore_default_theme: bool,
+
+    removed_entities: HashSet<Entity>,
```
geom3trik, line 92: The `EntityManager` in context actually already keeps track of which entities are 'alive'. You should be able to query it with the entity using the `is_alive` method.

geom3trik (commented):
Updating the observers in tree order is something I've wanted to do for a while but didn't realise it was so simple to implement.

I'm wondering also if this fixes a bug I found recently with the crud example. If you run that example and press the delete button then it crashes. I haven't looked into it yet but I think it might be an order of updates issue.

core/src/context/mod.rs
rhelmot: But entity IDs are recycled, aren't they? How can we tell the difference between an entity that was never destroyed and an entity that was destroyed and a new one built in its place?

geom3trik: Entities have a generation part to them which is tracked by the id manager. So the entity id might be the same but the generation will be different.

geom3trik (approved):

PR #194, opened 2022-08-09
Refactor systems into module

geom3trik:


core/src/context/backend.rs
```diff
@@ -360,29 +261,31 @@ impl<'a> BackendContext<'a> {
         // Not ideal
         let tree = self.0.tree.clone();
 
-        apply_inline_inheritance(self.0, &tree);
+        inline_inheritance_system(self.0, &tree);
 
         if self.0.style.needs_restyle {
-            apply_styles(self.0, &tree);
+            style_system(self.0, &tree);
```
rhelmot, line 267: Will the borrow checker let us move the check and the needs_restyle thing into the system call? it would be nice to make this just a list of calls.

core/src/context/backend.rs
```diff
@@ -360,29 +261,31 @@ impl<'a> BackendContext<'a> {
...
 
-        apply_z_ordering(self.0, &tree);
-        apply_visibility(self.0, &tree);
+        z_ordering_system(self.0, &tree);
+        visibility_system(self.0, &tree);
 
         // Layout
         if self.0.style.needs_relayout {
-            apply_text_constraints(self.0, &tree);
+            text_constraints_system(self.0, &tree);
```
rhelmot, line 288: as per above

rhelmot (approved):

rhelmot (approved):

Issue #171, opened 2022-06-24
Move things out of `core` into their own crate


KDecay:
We should consider moving things from `core` into their own crate (in the same repository) if reasonable.

# Possible candidates

- [ ] Storage types (`SparseSet`, `AnimatableSet`, `StyleSet`)
- [ ] Style system (probably when integrating the new [style system](https://github.com/vizia/style_system))
- [ ] Views

PR #196, opened 2022-08-11
Add `vizia_storage` and `vizia_id`

KDecay:
# Objective

- Progress towards #171

# Solution

- Add `vizia_storage` crate and move the `SparseSet` there.
- Add `vizia_id` and move the `GenerationalId` and `IdManager` there to be able to depend on it in `vizia_core` and `vizia_storage` to avoid a circular dependency between `vizia_core` and `vizia_storage`.
- Renamed `DenseIndex` to `SparseSetIndex`.
- Switched the generic index and value of the `SparseSetGeneric` to fit the definition of `HashMap`. Currently the index has the generic `I` which should be renamed to `K` once we can get rid of the `K: GenerationalId` on the `SparseSetGeneric` methods.

# Realizations during this PR

- The `SparseSetIndex` and `GenerationalId` are extremely similar but currently both are required. We should most likely figure out a way to combine them to avoid redundancy.
- The fields of the `SparseSet` are public which is required because they are used in the `StyleSet` for example. We should try to make these fields private because modifying them is the task of the `SparseSet`. It would also ensure that the sparse and dense vectors are modified correctly.
- `vizia_id` is a side effect of the effort of moving `SparseSetGeneric` to `vizia_storage`. During the move I realized that a lot of things are extremely dependant on each other which makes the splitting process a lot harder. I think moving these out into very simple and small crates is fine for now and if we feel like they could be combined with something else we can do that afterwards. This currently helps a lot to make the splitting proccess easier.


rhelmot:
A note: as we add more crates, we should update the advice for which crates you need to enable optimizations on for debug builds in order to ensure tolerable performance, and we should document that advice somewhere.

geom3trik (approved):

PR #197, opened 2022-08-13
Fix bugs with textbox introduced by #187

geom3trik:
Fix bugs with textbox introduced by #187.

- The textbox can remain focused but not in an edit state. The `:check` pseudoselector is used to show the cursor and border when the textbox is in an edit state.

- The `TriggerDown` event was not being sent to the captured entity which meant that the textbox could not be set to non edit by clicking away from it.

- The `Enter` key used for the Trigger event was not working correctly with the submit event of the textbox which is also triggered with the `Enter` key so I removed the `Enter` key causing a Trigger event.

### DIFF
--- a/crates/vizia_core/resources/themes/default_theme.css
+++ b/crates/vizia_core/resources/themes/default_theme.css
@@ -133,11 +133,10 @@ textbox:hover {
     background-color: #f6f6f6;
 }
 
-textbox:focus {
+textbox:checked {
     border-color: #4c00ff;
 }
-
-textbox:focus .textbox_content {
+textbox:checked .textbox_content {
     caret-color: #ff0000;
     selection-color: #6464c888;
 }
--- a/crates/vizia_core/src/context/backend.rs
+++ b/crates/vizia_core/src/context/backend.rs
@@ -360,9 +360,13 @@ impl<'a> BackendContext<'a> {
                         {
                             self.with_current(self.0.hovered, |cx| cx.focus_with_visibility(false));
                         }
-                        self.with_current(self.0.hovered, |cx| {
-                            cx.emit(WindowEvent::TriggerDown { mouse: true })
-                        });
+
+                        self.dispatch_direct_or_up(
+                            WindowEvent::TriggerDown { mouse: true },
+                            self.0.captured,
+                            self.0.triggered,
+                            true,
+                        );
                     }
                     MouseButton::Right => {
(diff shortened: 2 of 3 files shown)


geom3trik:
@0HyperCube would you mind taking a look at this? Just want to make sure I'm not messing something up with your focus/trigger event changes.

0HyperCube:
> The textbox can remain focused but not in an edit state

This is not what html does - if you click on a non-focusable widget then it focuses the root.

> The TriggerDown event was not being sent to the captured entity which meant that the textbox could not be set to non edit by clicking away from it.

Good spot

> The Enter key used for the Trigger event was not working correctly with the submit event of the textbox which is also triggered with the Enter key so I removed the Enter key causing a Trigger event.

I think that enter should continue to cause trigger down (it still also sends a keydown event). Perhaps the conflict was that the textbox is using the trigger events for selection rather than the mouse events?

geom3trik:
> This is not what html does - if you click on a non-focusable widget then it focuses the root.

What about when the user presses the enter key after editing the textbox? The textbox is no longer editable (until the user clicks on it again) but should it remain focused?

> I think that enter should continue to cause trigger down (it still also sends a keydown event). Perhaps the conflict was that the textbox is using the trigger events for selection rather than the mouse events?

The issue I was having was that with a list of textboxes (like the textbox_list example) you could tab between them fine but then pressing the enter key to submit the text would cause a different textbox to gain focus and enter edit mode. I'll admit I did take the lazy approach to a solution so I'll take a another look and try to keep enter as a trigger key.

geom3trik:
Going to merge this for now so we at least have a working textbox again and then fix the remaining issues in a follow-up PR.

0HyperCube:
> What about when the user presses the enter key after editing the textbox? The textbox is no longer editable (until the user clicks on it again) but should it remain focused?

The textbox is still editable for me after enter in the demo in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

geom3trik:
> The textbox is still editable for me after enter in the demo in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

Ah okay that's interesting. What do  you think would be the correct way to hadle the situation where the user does want the textbox to de-activate (but remain focused) when the enter key is pressed? Because I've definitely seen this behaviour before. Perhaps sending an event in the `on_submit` callback of the textbox?

PR #201, opened 2022-08-15
Restore trigger with enter

0HyperCube:
Restore trigger with enter whilst keeping the origional behaviour of being able to disable editing with enter.


geom3trik (approved):
Looks good to me!

PR #202, opened 2022-08-15
Add arrow key support for knob

0HyperCube:
Arrow key support for changing knob. Part of #55.

### DIFF
--- a/crates/vizia_core/resources/themes/default_theme.css
+++ b/crates/vizia_core/resources/themes/default_theme.css
@@ -283,6 +283,11 @@ dropdown vstack {
     child-space: auto;
 }
 
+knob:focus-visible {
+    outline-width: 2px;
+    outline-color: blue;
+}
+
 slider {
     background-color: #868686;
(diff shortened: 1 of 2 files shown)


geom3trik (commented):
I've noticed with this one that the tabbing to the knob, to give it the the `:focus-visible` style, and then clicking on the second knob gives the second one the ficus visible style. This doesn't seem to be consistent with what happens with the buttons or the sliders in #203.

Also, this might not be specific to this PR, but is it possible to remove the focus-visible style when clicking off of the control? Looking at what html does, it seems that focusing a control with tab gives it the focus-visible style but then clicking away removes it.

0HyperCube:
> Also, this might not be specific to this PR, but is it possible to remove the focus-visible style when clicking off of the control? Looking at what html does, it seems that focusing a control with tab gives it the focus-visible style but then clicking away removes it.

I added this to #201 but forgot to mention it.

geom3trik (approved):

PR #203, opened 2022-08-15
Add arrow key support for Slider

0HyperCube:
Arrow key support for changing sliders. Part of #55.

### DIFF
--- a/crates/vizia_core/src/views/slider.rs
+++ b/crates/vizia_core/src/views/slider.rs
@@ -1,4 +1,4 @@
-use std::{marker::PhantomData, ops::Range};
+use std::ops::Range;
 
 use crate::prelude::*;
 use crate::views::Orientation;
@@ -203,7 +203,7 @@ impl<L: Lens<Target = f32>> View for Slider<L> {
             }
 
             SliderEventInternal::SetKeyboardFraction(keyboard_fraction) => {
-                self.internal.keyboard_fraction = keyboard_fraction;
+                self.internal.keyboard_fraction = *keyboard_fraction;
             }
         });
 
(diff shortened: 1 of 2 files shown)


crates/vizia_core/src/views/slider.rs
```diff
@@ -97,14 +99,15 @@ where
     /// ```
     pub fn new(cx: &mut Context, lens: L) -> Handle<Self> {
         Self {
-            p: PhantomData::default(),
+            lens: lens.clone(),
```
geom3trik, line 102: Since `PhantomData` isn't being used any more can you remove the import to remove the warning.

crates/vizia_core/src/views/slider.rs
```diff
@@ -197,6 +201,10 @@ impl<L: Lens> View for Slider<L> {
             SliderEventInternal::SetRange(range) => {
                 self.internal.range = range.clone();
             }
+
+            SliderEventInternal::SetKeyboardFraction(keyboard_fraction) => {
+                self.internal.keyboard_fraction = keyboard_fraction;
```
geom3trik, line 206: Compiler is complaining that `keyboard_fraction` needs dereferencing here.

geom3trik (changes requested):
Looks good! Just a couple of minor things stopping it from compiling at the moment.

crates/vizia_core/src/views/slider.rs
0HyperCube: Wasn't an error on whatever version of rust I used to be using but it is on the current stable.

geom3trik (approved):

PR #206, opened 2022-08-16
Improvements to debug tree (i key)

0HyperCube:
- The tree structure is more clear (using ascii art from the `tree` command)
- Bindings are labelled with what they are binding to
- All views in core are labelled

### DIFF
--- a/crates/vizia_core/src/state/binding.rs
+++ b/crates/vizia_core/src/state/binding.rs
@@ -4,30 +4,6 @@ use std::collections::{HashMap, HashSet};
 use crate::prelude::*;
 use crate::state::{BasicStore, LensCache, ModelOrView, Store, StoreId};
 
-/// A view on a binding to allow for better debugging
-struct BindingView<L: Lens> {
-    lens: L,
-}
-
-impl<L: Lens> View for BindingView<L> {
-    fn build<F>(self, cx: &mut Context, _content: F) -> Handle<Self> {
-        let id = cx.entity_manager.create();
-        let current = cx.current();
-        cx.tree.add(id, current).expect("Failed to add to tree");
-        cx.cache.add(id).expect("Failed to add to cache");
-        cx.style.add(id);
-        cx.views.insert(id, Box::new(self));
-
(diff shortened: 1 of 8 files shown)


geom3trik (commented):
This is really good, thank you. I have a couple of ideas, just want to get your opinion:

- I think the trigger key should be something like CTRL + I, so it isn't triggered while testing something like a textbox.
- I see that bindings are currently displayed as 'no view'. I think it would be good to label these as bindings.

For that second point, I think it might be cool if there was some way to display what the binding was to. But this would involve requiring lenses to be labelled. It's possible we could maybe adjust the lens derive macro to generate a function which returns the name of the lens source/target. But this is probably out of scope for this PR. So for now I think 'binding' instead of 'no view'.

0HyperCube:
Added the <kbd>ctrl</kbd> + <kbd>i</kbd> and the binding names.

crates/vizia_core/src/state/binding.rs
```diff
@@ -4,6 +4,30 @@ use std::collections::{HashMap, HashSet};
...
+        Handle { entity: id, p: Default::default(), cx }
+    }
+
+    fn element(&self) -> Option<&'static str> {
+        self.lens.name()
+    }
+
+    fn draw(&self, _cx: &mut DrawContext, _canvas: &mut Canvas) {}
+}
+
```
geom3trik, line 30: I don't think we should have a dedicated view for bindings. We had that before but it adds superfluous information, such as a bounding box and clip region which are meaningless for bindings.

It should be possible to get the lens name from the binding if you add a method to `BindingHandler`, because all of the bindings are stored in hashmap in context with the entity as the key.

geom3trik (changes requested):

crates/vizia_core/src/state/binding.rs
0HyperCube: Ok

geom3trik (approved):
Awesome, love this change btw. Thank you again!

PR #208, opened 2022-08-17
Add APIs to adjust some femtovg cache sizes

rhelmot:
These are necessary for arborio to function performantly :) I was previously accessing text_context directly off of Context, but that was made private in the BackendContext cleanup.


geom3trik (approved):
Looks good to me. Maybe if we end up with a lot of these configuration type things we might want to consider a `with_config()` method and putting them all into one struct, but for now I'm happy with this.

PR #209, opened 2022-08-17
Textbox improvements

0HyperCube:
- Double click to select word and triple click to select paragraph
- When editing is false, you can now properly select a range by dragging. If you don't drag, then it will select all as before.
- When you click outside a textbox, it didn't send a triggerdown event, only a mousedown.

### DIFF
--- a/crates/vizia_core/src/context/backend.rs
+++ b/crates/vizia_core/src/context/backend.rs
@@ -378,7 +378,7 @@ impl<'a> BackendContext<'a> {
                     if self.0.clicks <= 2 {
                         self.0.clicks += 1;
                         let event = if self.0.clicks == 3 {
-                            WindowEvent::MouseTrippleClick(button)
+                            WindowEvent::MouseTripleClick(button)
                         } else {
                             WindowEvent::MouseDoubleClick(button)
                         };
--- a/crates/vizia_core/src/context/backend.rs
+++ b/crates/vizia_core/src/context/backend.rs
@@ -321,7 +321,7 @@ impl<'a> BackendContext<'a> {
 
                 self.dispatch_direct_or_up(event, self.0.captured, self.0.hovered, false);
             }
-            WindowEvent::MouseDown(button) => {
+            &WindowEvent::MouseDown(button) => {
                 match button {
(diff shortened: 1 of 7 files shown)


geom3trik:
These are good additions but I've noticed some unexpeted behaviour.

From the textbox example:
1. Double clicking after the text selects the last word but excludes the exclamation mark.
2. Clicking once in one place and then clicking quickly in another selects the word under the second click. This is odd because it looks like the check for the second click being in the same place as the first is there so I'm not sure why this is happening.

0HyperCube:
1. I've improved the word selection to match gedit.
2. I was accidentally setting `self.0.click_pos = new_click_pos` before the comparison ;).

geom3trik:
Looks like those bugs are fixed but I'm afraid I've found another one. If you mouse down outside of the textbox (before it's active) and then drag into it and release the mouse button, it selects the text. I *think* you just need a check on mouse up to make sure that the mouse down happened on the textbox.

0HyperCube:
Indeed, I just moved the start edit into the mouse up without thinking about dragging.

geom3trik:
Okay, textbox example works great now. However, I think I've found another bug on the `textbox_list` example. Clicking between the different textboxes no longer selects all the text. It's a bit inconsistent, sometimes it does and sometimes it doesn't.

0HyperCube:
The focus in event was conflicting with the mouse down - I just had to check that the mouse is not pressing current when focus in occurs.

geom3trik:
> The focus in event was conflicting with the mouse down - I just had to check that the mouse is not pressing current when focus in occurs.

Yep that's fixed that problem, but I'm afraid I've found another (maybe related?) problem. On the `textbox_list` example again, if you click and release on the first textbox to focus it, then click and drag across the second (or any other) textbox until the cursor is outside of it, then release the mouse button, the cursor icon remains on the text one and clicking outside of the textbox doesn't defocus it. Also, while dragging the cursor of the textbox, it doesn't work if your cursor leaves the bounds of the textbox when it still should (above at least, below is broken but that's a known issue).

0HyperCube:
I think that I've fixed both of those issues?

geom3trik (approved):

PR #212, opened 2022-08-19
Update to newest femtovg/glutin/winit; fix clipboard on wayland

### LINKED ISSUE #113: Move clipboard to window for Wayland support
As mentioned in #110 

> Fully implementing Wayland support will take more work as we need need a
RawWidnowHandle to request a proxy, which is not available until after
the Context is created.

The clipboard proxy should be created at the point that the window is created and the window handle is available rather than when the context is created. This requires either moving the clipboard handle to the window, or we can place the clipboard behind an `Option` in context.

rhelmot:
Closes #113 

We should pin to femtovg 0.3.6 whenever it is released.


rhelmot:
This PR now reverts part of the crates refactor - run-wasm needs to be in the repository root, otherwise it will create a `target` dir under `crates`, which prevents further builds from happening.

rhelmot:
I'm not entirely sure what it'll take to support clipboard+wayland+baseview. I think a feature is missing in baseview to support getting the pointer necessary to initialize the clipboard context?

geom3trik:
> I'm not entirely sure what it'll take to support clipboard+wayland+baseview. I think a feature is missing in baseview to support getting the pointer necessary to initialize the clipboard context?

Last time I checked baseview didn't support wayland at all. I think it's just Windows, MacOS, and x11.

rhelmot:
I was able to run the sample applications under wayland+baseview :)

But if it's not meant to be in scope, then this can be merged.

geom3trik:
> I was able to run the sample applications under wayland+baseview :)
> 
> But if it's not meant to be in scope, then this can be merged.

I checked with the devs and apparently baseview does not support wayland as I thought, so when you run the examples you're using xwayland. As such there is no way to get a wayland display pointer for the clipboard.

In which case I'm happy to merge this now if you are.

geom3trik (approved):

PR #215, opened 2022-08-23
Switch view

LunaeSomnia:
A really basic switch view, created from the checkbox code.

Also added two different stylesheets named `dark_theme.css` and `light_theme.css` under `crates/vizia_core/resources/themes/` as they will be needed in a future.


crates/vizia_core/src/views/switch.rs
```diff
@@ -0,0 +1,205 @@
...
+/// #
+/// # let cx = &mut Context::new();
+/// #
+/// # AppData { value: false }.build(cx);
+/// # use vizia_core::fonts::unicode_names::CANCEL;
+///
+/// Switch::new(cx, AppData::value)
+///     .on_toggle(|cx| cx.emit(AppEvent::ToggleValue))
+///     .text(AppData::value.map(|flag| if *flag {CANCEL} else {""}));
+/// ```
```
geom3trik, line 109: This doesn't apply to the switch and can be removed.

crates/vizia_core/src/views/switch.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// # AppData { value: false }.build(cx);
+    /// #
+    /// Switch::new(cx, AppData::value);
+    /// ```
+    pub fn new(cx: &mut Context, checked: impl Lens<Target = bool>) -> Handle<Self> {
+        //let checked = checked.get_val_fallible(cx).unwrap_or(false);
+        Self { on_toggle: None }
+            .build(cx, |cx| {
+                HStack::new(cx, |cx| {
+                    HStack::new(cx, |_|{}).class("switch-handle").hoverable(false);
```
geom3trik, line 140: You can use an `Element` here instead of a `HStack` because it doesn't need to be a container.

crates/vizia_core/src/views/switch.rs
```diff
@@ -0,0 +1,205 @@
...
+                }).class("switch-handle-container").hoverable(false);
+            })
+            .bind(checked, |handle, checked| {
+                if let Some(flag) = checked.get_val_fallible(handle.cx) {
+                    handle.checked(flag);
+                }
+            })
```
geom3trik, line 147: I think in this case you should be able to pass the lens directly to the `checked` modifier.

crates/vizia_core/src/views/switch.rs
LunaeSomnia: Okay, I'll remove all coments as those can be written in the future with no issues

geom3trik: Oh, it's only the highlighted section that needs to be removed (because the switch doesn't use text to display the inner part). The rest of the docs are fine.

LunaeSomnia: Okay. Changed!

.vscode/settings.json
```diff
@@ -2,11 +2,11 @@
 	"[rust]": {
 		"editor.formatOnSave": true,
 		"editor.formatOnPaste": true,
-		"editor.defaultFormatter": "rust-lang.rust-analyzer",
+		"editor.defaultFormatter": "matklad.rust-analyzer",
```
0HyperCube, line 5: [Rust Analyser is now part of the rust-lang org](https://blog.rust-lang.org/2022/02/21/rust-analyzer-joins-rust-org.html). If you are still using the old extension you should consider updating.

LunaeSomnia: Oh, thanks for letting me know. I didn't mean to change that...

LunaeSomnia:
Moved to PR #217

PR #216, opened 2022-08-28
Spinbox View

LunaeSomnia:
Created a spinbox view that will be used to create other widgets, like the `Datepicker`.

image

image


LunaeSomnia:
I just found a bug that I don't know exactly how to solve. On the example I built, there are two different types of data that should behave separately but currently all of the events go to one of the types of `Spinbox` (the most right one). It is 100% related on how I built the children on the tree, but I lack the expertise to debug that.

LunaeSomnia:
Moved to PR #218

PR #223, opened 2022-08-30
Add Rays and Event::take

rhelmot:
A Ray is a new concept which is just a value pending application to a struct. Deriving Ray on `MyStruct` produces a `MyStructRay` enum, which has a variant for each field of the struct. You can call Ray::apply or Ray::swap to apply a ray to the relevant struct.

I've been programming with a prototype of this paradigm in arborio for a few weeks, and the thing I found is that sometimes you want the strike operation to actually be a swap between the struct and the enum - this allows you to inspect the result of the operation post-facto and also gets you an inverse of the event for free - perfect for implementing undo-redo. In order to facilitate this, I had to put RefCells in all of my events, so to smooth that over I introduced Event::take. This required getting rid of the Message type because [mumble mumble something vtables]. Everywhere that previously said Message now says Any + Send.

### DIFF
--- a/examples/views/textbox.rs
+++ b/examples/views/textbox.rs
@@ -1,14 +1,8 @@
 use vizia::prelude::*;
 
-#[derive(Lens, Ray)]
+#[derive(Lens, Setter, Model)]
 pub struct AppData {
     text: String,
-}
-
-impl Model for AppData {
-    fn event(&mut self, _: &mut EventContext, event: &mut Event) {
-        event.take::<AppDataRay>().map(|ray| ray.apply(self));
-    }
 }
 
 fn main() {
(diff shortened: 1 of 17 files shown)


examples/views/textbox.rs
```diff
@@ -1,22 +1,13 @@
...
-}
-
 impl Model for AppData {
     fn event(&mut self, _: &mut EventContext, event: &mut Event) {
-        event.map(|app_event, _| match app_event {
-            AppEvent::SetText(text) => {
-                self.text = text.clone();
-            }
-        });
+        event.take::<AppDataRay>().map(|ray| ray.apply(self));
```
geom3trik, line 10: Should we use this here or make a dedicated example for it?

geom3trik (approved):

examples/views/textbox.rs
rhelmot: So it depends on the way we want to structure our teaching. I think we should make it so that we introduce rays as a basic building block, after explaining how to do the same thing with "standard" events/messages.

rhelmot: Which is to say, I think we should convert examples to using rays as applicable and as we come across them

geom3trik: Okay, seems reasonable to me.

PR #231, opened 2022-09-10
Fix clipping problems

geom3trik:
I noticed that views couldn't be hovered and wouldn't be drawn outside the bounds of a transformed parent even if the overflow was set to visible. This is because overflow: visible would cause the view to clip to the root clip region (window bounds). This was actually 2 separate problems:

1. The hover system correctly transforms the view but if the view is moved outside of the bounds of the window then it can't be hovered, even though it may be visible thanks to the transform. To fix this I changed the root clip_region.
2. The draw system was ulling views outside of the window but not taking into account any transforms. To fix this the bounds are now transformed before culling.


geom3trik:
Seems that my solution does not work. Converting to draft until I figure this one out.

PR #236, opened 2022-09-22
Set modifiers on baseview mouse events

robbert-vdh:
Using the changes from https://github.com/RustAudio/baseview/pull/117.

Until baseview implements proper keyboard focus handling this helps when dragging a knob or slider outside of the window. Otherwise the modifier state may get stuck on an incorrect value.


rhelmot (approved):

rhelmot:
lgtm. Thanks!

PR #240, opened 2022-09-26
Use `Path` and `PathBuf` for stylesheets

geom3trik:


crates/vizia_core/src/context/mod.rs
```diff
@@ -406,9 +407,9 @@ impl Context {
         }
     }
 
-    pub fn add_stylesheet(&mut self, path: &str) -> Result<(), std::io::Error> {
-        let style_string = std::fs::read_to_string(path)?;
-        self.resource_manager.stylesheets.push(path.to_owned());
+    pub fn add_stylesheet(&mut self, path: impl AsRef<Path>) -> Result<(), std::io::Error> {
+        let style_string = std::fs::read_to_string(path.as_ref())?;
```
FreeFull, line 411: The .as_ref() shouldn't be needed on this line, because `read_to_string` also takes a `T: AsRef<Path>`

rhelmot (approved):

PR #245, opened 2022-09-28
fix on_release never happening

Fredemus:
Changing `MouseEnter` [here](https://github.com/vizia/vizia/blob/main/crates/vizia_core/src/modifiers/actions.rs#L163) to `MouseOver` seems to fix `on_hover` too, but from reading what it should do, I'd think something's wrong in [hover.rs](https://github.com/vizia/vizia/blob/main/crates/vizia_core/src/systems/hover.rs).

### DIFF
--- a/crates/vizia_core/src/modifiers/actions.rs
+++ b/crates/vizia_core/src/modifiers/actions.rs
@@ -102,7 +102,7 @@ impl<V: View> View for Release<V> {
         self.view.event(cx, event);
 
         event.map(|window_event, meta| match window_event {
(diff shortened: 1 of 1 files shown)


geom3trik:
While this may be a valid fix, I'm going to hold off on merging this because I have a change in the works that would replace this.

Fredemus:
Fair, just thought I'd see if I could find the problem and maybe offer a temporary fix. Do you want me to close this PR?

geom3trik:
Closing in favour of #180

PR #248, opened 2022-10-08
Add groundwork for runtime window resizing and user scales in baseview

robbert-vdh:
This is cherry picked and squashed from my patched vizia branch. This patch introduces the concept of a user scale and adds two new properties to the context: that user scale, and a window size field. The user scale makes it possible to uniformly resize a window at runtime independently of system HiDPI scaling. As a result, the setting can be saved and recalled separately in an application, and moving the window between displays with different HiDPI scaling factors preserves the user scale. The window size field meanwhile can be used to change the window's width and/or height without scaling the window's contents.

A consequence of this is that the DPI factor is now a combination of the window's DPI factor and the user scale factor. The scale factor fields in the baseview backend have thus been updated accordingly to make this clear. The system scaling policy is also taken into account here.

The workflow for uniformly rescaling the window is as follows:

1. A widget uses `cx.set_user_scale_factor(...)` to set a new user scale factor. Storing this directly on the context instead of sending events has the advantage that you get debouncing for free (there can be at most a single scale factor change per frame), setting the scale factor to the already current scale factor is a no-op, and other widgets can't react to the event and potentially do the wrong thing if the event needs to be thrown away.
2. After handling the events and just before drawing, the baseview backend checks if the user scale factor and window size fields have changed by comparing the old known values to the current values stored in the context.
3. If the values have changed, then the backend will resize the baseview window (this callback is currently missing).
4. Finally, the backend emits a `WindowEvent::WindowResize` to indicate that the window has been successfully resized. This event technically already existed, but it was not used anywhere.

Changing the window size using `cx.set_window_size(...)` works the same way.

What is missing:

- An example
- Widgets that allow resizing the window using this new API. I have a resize handle implementation [here](https://github.com/robbert-vdh/nih-plug/blob/master/nih_plug_vizia/src/widgets/resize_handle.rs). Another common option would be to have a dropdown menu with predefined sizes.
- Winit support (would require winit pathces)
- A finished implementation of the feature in baseview. Right now I have a branch that implements this on Linux. I still need to add Windows and macOS support, so the actual resize call in this patch has been commented out.

These patches have been around for a while and they've been refactored through at least half a dozen large changes of `vizia_baseview`, so it may not be entirely idiomatic anymore.


geom3trik:
This looks good to me in principle. When we have multiwindow the API will likely need to change but since that's an unknown amount of time away we can carry on for now. I'm curious what patches would be required for winit since it should already be possible to change the window size at runtime?

I'm also not sure about setting the window size directly through context. Window properties are currently set using events, there's also one for size `WindowEvent::SetSize`, but I'm wondering if this would work here or if it *has* to be set through context because of that user scale?

robbert-vdh:
> This looks good to me in principle. When we have multiwindow the API will likely need to change but since that's an unknown amount of time away we can carry on for now. I'm curious what patches would be required for winit since it should already be possible to change the window size at runtime?

Is there a way to change the window size at runtime? I couldn't find it when I looked into it, hence why the winit backend just doesn't do anything with these new fields.

> I'm also not sure about setting the window size directly through context. Window properties are currently set using events, there's also one for size WindowEvent::SetSize, but I'm wondering if this would work here or if it has to be set through context because of that user scale?

I explained before in the Discord why I think using context properties is a better solution for this problem, and I repeated some of those reasons in the OP here. Setting the properties using events will only add more moving pieces without making things any simpler (you'd then need to manually debounce events/take the last one, and you still need to keep track of the current/old value to ignore resizes that should be a no-op).

geom3trik:
> Is there a way to change the window size at runtime? I couldn't find it when I looked into it, hence why the winit backend just doesn't do anything with these new fields.

I thought there was but I'll have to double check that.

> I explained before in the Discord why I think using context properties is a better solution for this problem, and I repeated some of those reasons in the OP here. Setting the properties using events will only add more moving pieces without making things any simpler (you'd then need to manually debounce events/take the last one, and you still need to keep track of the current/old value to ignore resizes that should be a no-op).

The reason I'm tempted to keep them as events is for consistency of the API, where other properties are set with events. How expensive is it to change the window size? If we just applied multiple resize events in the same cycle or didn't keep track of old values what's the worst case scenario? I can't imagine it's too expensive and I don't imagine it's something that happens often but I'm happy to be proven wrong. Another reason for using events would be for when we eventually have multiwindow. An event can propagate up to the parent window of whatever emitted it but how do we determine the correct window when calling methods on `cx`? I guess context could probably figure out the parent window from the current entity. Just thinking out loud really.

robbert-vdh:
> How expensive is it to change the window size? If we just applied multiple resize events in the same cycle or didn't keep track of old values what's the worst case scenario?

It probably won't happen often, but in the context of plugins unnecessary resize calls can be pretty expensive (and in other cases it might result in some flickering if two widgets try to perform conflicting resizes). But yes a field indeed won't scale to multiple windows.

geom3trik:
> It probably won't happen often, but in the context of plugins unnecessary resize calls can be pretty expensive (and in other cases it might result in some flickering if two widgets try to perform conflicting resizes). But yes a field indeed won't scale to multiple windows.

Would there be flickering if the two widgets try to resize within the same frame? I would have thought not.

I do see the advantage to de-duplicating these events though. Since multiwindow doesn't apply to the baseview backend this might be place where the API could diverge, but ideally it would be good to have the same thing for both windowing backends.

robbert-vdh:
Any updates on this? If you want I can rewrite it to a version that uses events and that doesn't have a confirmation of the resize (`WindowEvent::WindowResize` will be gone/unused again) and no debouncing.

geom3trik:
> Any updates on this? If you want I can rewrite it to a version that uses events and that doesn't have a confirmation of the resize (`WindowEvent::WindowResize` will be gone/unused again) and no debouncing.

Sorry for the delay. I think until a mechanism is in place for de-duplicating events (which I'm not currently working on) there's probably no point switching to an event-based approach. When we eventually have multiwindow we might rethink how we do this but for now I'm happy with direct mutation of the context. I'll do a proper review of the code hopefully later today but the first think I noticed glancing at it again was the change to `Context::new`. I think we should in that case have a `Context::default` so the doctests don't look so messy.

geom3trik:
I would prefer it that way generally, to keep it consistent with other window modifications and to make things easier for when we have multiwindow, but at this stage I'm not bothered either way. Whichever is easier for you really. An event would cause an unnecessary redraw but I doubt you'd notice it with the baseview backend, which is redrawing every update cycle anyway (for now).

robbert-vdh:
Btw I also finally implemented resizing in baseview for Windows (that was a major pain compared to how simple it was to add for Linux heh). I'll do macOS next and then submit a PR to the baseview repo.

robbert-vdh:
https://github.com/RustAudio/baseview/pull/136 was merged a couple days ago, so I'll update this to use the actual resize method. It would be great to have this merged since there are already a bunch of conflicts again. I've spent at least a dozen hours resolving these conflicts over the last year. :sweat_smile:

geom3trik:
> [RustAudio/baseview#136](https://github.com/RustAudio/baseview/pull/136) was merged a couple days ago, so I'll update this to use the actual resize method. It would be great to have this merged since there are already a bunch of conflicts again. I've spent at least a dozen hours resolving these conflicts over the last year. 😅

Is there still the bug we identified before? Where the window doesn't resize correctly and we suspected was due to vizia rather than baseview?

robbert-vdh:
I haven't had time to look at that yet (nor do I have any way to debug it). Your issue only seemed to happen on Windows with DPI scaling.I f I force DPI scaling on Linux everything is fine, and the straight up resizing the window also seemed to work fine on Windows regardless of HiDPI scaling.

geom3trik:
> I haven't had time to look at that yet (nor do I have any way to debug it). Your issue only seemed to happen on Windows with DPI scaling.I f I force DPI scaling on Linux everything is fine, and the straight up resizing the window also seemed to work fine on Windows regardless of HiDPI scaling.

What's the best way to test the changes in this PR? Could you perhaps add an example with a drag handle like the one in your plugins? Doesn't have to be fancy or anything.

robbert-vdh:
I've been working on updating this and on an example where you can resize the window using both a text field and a resize handle. Two observations so far:

- I can change the setting of the user scale and window size to an event, _but_ this information still needs to be stored on the context to be able to access it. I'll do this after I'm done with the rest of the changes.
- I initially repurposed the unused `WindowEvent::WindowResize` to inform the application that its size has changed, but I realize that this is not actually needed. `WindowEvent::GeometryChanged` already covers this. Should I just get rid of `WindowEvent::WindowResize`?

robbert-vdh:
I've pushed what I have for now including the partial example. Some additional observations:

- I still need to add the resize handle to the example. Right now you can only resize the window by typing in a new scale value. It should probably also have a way to change the window's logical size (which is also supported by this PR).
- The `WindowEvent::GeometryChanged` seems to take a second to fire the first time. This could just be debug builds being slow though. After the first time it's instantaneous and this delay also doesn't exist for release builds.
- Something seems to clamp the standalone window's size to the display's size, at least on Linux. As a result setting the scale factor to a value that would cause the window to be clipped and then setting it to 1.0 again will make the window smaller than its original size.

geom3trik (approved):

PR #258, opened 2022-11-12
Add `EventHandle` so that action modifiers can mutate view data

geom3trik:
While working with action modifiers, e.g. `on_press(|cx| ...)` I realised it would be very useful to be able to mutate the actual view data. To do this the view type needs to be known so that the view can be retrieved from the context. This PR adds an `EventHandle`, which is similar to `Handle` but wraps an `EventContext`. This is then used by action modifiers to provide a way to mutate the view data in response to an event. See `action_modifier` example to see how this works.

### DIFF
--- a/crates/vizia_core/src/context/event.rs
+++ b/crates/vizia_core/src/context/event.rs
@@ -32,7 +32,7 @@ pub struct EventContext<'a> {
     pub draw_cache: &'a DrawCache,
     pub tree: &'a Tree<Entity>,
     pub(crate) data: &'a SparseSet<ModelDataStore>,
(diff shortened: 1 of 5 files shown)


rhelmot (approved):

rhelmot:
This is pretty cool :)

geom3trik:
One question I have now is whether this `EventHandle` should be used in place of `EventContext` for all callbacks? Or should it be exlusively for the action modifiers?

rhelmot:
I think as long as EventHandle duck-types as EventContext we're fine.

PR #259, opened 2022-11-12
Make `Lens` `Copy` Again

geom3trik:
No more cloning lenses to move them into closures!!!

### DIFF
--- a/crates/vizia_core/src/localization/mod.rs
+++ b/crates/vizia_core/src/localization/mod.rs
@@ -38,7 +38,7 @@ where
         Box::new(self.clone())
     }
     fn bind(&self, cx: &mut Context, closure: Box<dyn Fn(&mut Context)>) {
-        Binding::new(cx, self.lens.clone(), move |cx, _| closure(cx));
+        Binding::new(cx, self.lens, move |cx, _| closure(cx));
     }
 }
 
--- a/crates/vizia_core/src/state/binding.rs
+++ b/crates/vizia_core/src/state/binding.rs
@@ -44,7 +44,7 @@ where
         cx.cache.add(id).expect("Failed to add to cache");
         cx.style.add(id);
 
-        let binding = Self { entity: id, lens: lens.clone(), content: Some(Box::new(builder)) };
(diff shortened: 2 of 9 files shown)


rhelmot:
This is very interesting... I was initially opposed to this change because I thought I had a bunch of lenses which held strings by value, but then I checked my implementations and it seems that all of those lenses actually use _another lens_ to get ahold of the string they need to use. I have some reservations about whether this pattern can hold, but I'm optimistic!

geom3trik:
> This is very interesting... I was initially opposed to this change because I thought I had a bunch of lenses which held strings by value, but then I checked my implementations and it seems that all of those lenses actually use _another lens_ to get ahold of the string they need to use. I have some reservations about whether this pattern can hold, but I'm optimistic!

I too have some reservations but if this can work then I think it's a big ergonomics improvement not having to clone lenses for nested closures. I tested this with the `map()` lens and it still allows you to map to a string even if it's not derived from the original data, e.g. `lens.map(|_| String::from("Test"))`. I think in general lenses shouldn't contain non-copy data and should instead try to refer to app data if it's non-copy or use a map. I've tested this so far with all of the built-in views and examples on the `new-views` branch and it seems to work. I have a couple of other more complex code bases which use vizia to test it with and will report back. It seems promising though that this works with your application if I'm understanding correctly?

rhelmot:
I have not tested that it works with my application, just reviewed my custom lenses and determined that they will probably port with no issues whatsoever.

geom3trik:
> I have not tested that it works with my application, just reviewed my custom lenses and determined that they will probably port with no issues whatsoever.

Ah okay fair enough. I'm optimistic that it will work then. I have an application I can test with but I'm not doing anything fancy with lenses so it might not be the best tester.

geom3trik:
After testing this further in a more realistic scenario, I've come to the conclusion that this change is too restrictive and not viable. It's a shame because I really would like to be able to reduce the cloning of lenses, but this isn't the solution. I'm going to close this for now but leave the branch in case it's something we can come back to in the future.

PR #263, opened 2022-11-17
Optimize draw call in view trait

KDecay:
The diff looks a bit wild, but I just moved the content of the `draw` function of the `View` trait into a separate function called `draw_view`. This reduces the binary size quite a bit.

# Comparison

### Dropdown Example

Old: 13752 kb
New: 13484 kb

### Crud Example

Old: 13936 kb
New: 13608 kb

### Dropdown Example

Old: 8784 kb
New: 8696 kb

### Crud Example

Old: 8972 kb
New: 8832 kb

### DIFF
--- a/crates/vizia_core/src/view.rs
+++ b/crates/vizia_core/src/view.rs
@@ -55,635 +55,7 @@ pub trait View: 'static + Sized {
     fn event(&mut self, cx: &mut EventContext, event: &mut Event) {}
 
     fn draw(&self, cx: &mut DrawContext, canvas: &mut Canvas) {
-        let bounds = cx.bounds();
-
-        //Skip widgets with no width or no height
-        if bounds.w == 0.0 || bounds.h == 0.0 {
-            return;
-        }
-
-        let background_color = cx.background_color().cloned().unwrap_or_default();
-
-        let font_color = cx.font_color().cloned().unwrap_or(Color::rgb(0, 0, 0));
-
-        let border_color = cx.border_color().cloned().unwrap_or_default();
-        let outline_color = cx.outline_color().cloned().unwrap_or_default();
-
-        let parent = cx
-            .tree
-            .get_layout_parent(cx.current)
-            .expect(&format!("Failed to find parent somehow: {}", cx.current));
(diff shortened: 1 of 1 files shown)


robbert-vdh:
Did you try comparing compile times with and without this change, particularly for slightly less trivial applications? By moving the drawing code to a function like this significantly less code needs to be generated, so I can imagine that optimizing may also become faster.

geom3trik (approved):

KDecay:
> Have you tried comparing compile times with and without this change, particularly for slightly less trivial applications? By moving the drawing code to a function like this significantly less code needs to be generated, so I can imagine that optimizing may also become faster.

@robbert-vdh I haven't yet but I'll put that on my todo list and let you know the results! I'm certain that it has a positive effect on compile time as well.

KDecay:
@robbert-vdh @geom3trik I just tested the compile time difference this change made and it's pretty huge. I made a simple crate that just defines 100 types and implements the default `View` implementation (`impl View for Foo {}`). I compiled it using `cargo build --timings` and checked the time it took the crate to build (just the crate not `Vizia`).

The results are as follows:
- Before: 6.95s
- After: 4.27s

Of course this is not a scientific result by any means but it should show how much impact this change had.

robbert-vdh:
I'm also curious about the binary size difference between the two (if you just add all those 100 views in an application so they don't get optimized out).

KDecay:
Alright I tested it again now using 1000 Types. All of the builds use `winit` and are build using the following commands:
- Release: `cargo build --release --timings`
- Debug: `cargo build --timings`

# Debug builds

## Before 

Time: 36.88s
Size: 237MB

## After

Time: 8.8s
Size: 169MB

# Release builds

## Before 

Time: 211.86s
Size: 27MB

## After

Time: 13.67s
Size: 11MB

geom3trik:
Wow, these time differences are insane! Is this a clean build both times? I.e. no incremental builds?

KDecay:
@geom3trik Yup I always ran `cargo clean` before every test.

PR #267, opened 2022-12-03
Massive knob changes

LunaeSomnia:


crates/vizia_core/src/views/knob.rs
```diff
@@ -10,143 +10,252 @@ static DEFAULT_MODIFIER_SCALAR: f32 = 0.04;
...
+#[derive(Lens)]
+pub struct Knob<L: Lens<Target = D>, D: Copy + Clone + Into<f32> + 'static> {
+    lens: L,
     is_dragging: bool,
     prev_drag_y: f32,
-    continuous_normal: f32,
+    cursor: D,
+    default_normal: D,
+    min_value: D,
+    max_value: D,
```
Fredemus, line 28: I don't love the idea of knobs having a mix and min value. In my experience it's easier to have that applied to whatever the knob is changing in an event, especially if the curve between min and max isn't linear

Fredemus (commented):
I like all of these changes, except for the knobs having min and max values. It makes them harder to use in the plugin case at least

crates/vizia_core/src/views/knob.rs
LunaeSomnia: Makes total sense. I haven't even considered non-linear while re-implementing the knob 😅 

I think a possible solution to this is not exposing the `min_value` and `max_value` from the constructor and have them fixed at `0.0` and `1.0` respectively for the continuous knob. 

How does that sound to you?

Fredemus: Yeah that sounds like a fine solution to me

geom3trik (approved):

geom3trik:
Just FYI I'm holding out on merging this until I've finished the accesskit PR because in that PR I had to make some changes to sliders to make them accessible and I'd like to do the same with the knob.

PR #269, opened 2022-12-05
Replace `Data` trait with `Clone + PartialEq`

### LINKED ISSUE #268: Remove the `Data` trait
Currently we are using the `Data` trait to compare mutated state with an stored 'old' value. The comparison used in the `Data` trait is supposed to be fast, which is why for things like `Arc` it does a pointer comparison rather than comparing the inner value. However, in practical use the `Data` trait is just used to directly compare, and wrapper types like `Arc` and `Mutex` are often mapped (using lens map) to get their inner value.

I suggest we remove the `Data` trait and instead use `PartialEq`. For cases where we need a fast comparison we could use a wrapper type of our own with a custom `PartialEq` implementation.

This change will remove quite a few restrictions when dealing with lensed data, and also remove the need to wrap types from 3rd party crates as long as they implement `PartialEq`.

Thoughts?

geom3trik:
Closes #268


rhelmot:
I was initially against this because I wanted to maintain the ability to have custom data definitions that e.g. compare just against a version number while maintaining the semantics of PartialEq, but then I realized that that is silly and you can just bind directly to the version number field.

rhelmot (approved):

geom3trik:
> I was initially against this because I wanted to maintain the ability to have custom data definitions that e.g. compare just against a version number while maintaining the semantics of PartialEq, but then I realized that that is silly and you can just bind directly to the version number field.

Could you not also manually implement `PartialEq` for a custom data type and do that?

rhelmot:
You could, but the ergonomics would suffer.

geom3trik:
> You could, but the ergonomics would suffer.

I suppose so. Then how does binding directly to a version number work? I guess you'd bind to a version number but use `cx.data()` to get the data you actually want? Since the binding will just give you the version number.

rhelmot:
Yeah - [here's](https://github.com/rhelmot/arborio/blob/78fe985eb0e8034ba262879e229a77a3a1b3be2a/crates/widgets/src/tabs/installation.rs#L24) an example of me doing that.

geom3trik:
> Yeah - [here's](https://github.com/rhelmot/arborio/blob/78fe985eb0e8034ba262879e229a77a3a1b3be2a/crates/widgets/src/tabs/installation.rs#L24) an example of me doing that.

Okay cool, that looks simple enough. Although, this would only really work for a `Binding` view and not so much for a property binding, though we do have `bind`  (on Handle) for that I guess. If we're happy with this switch from `Data` to `PartialEq` then maybe we can solve this particular problem with some special form of `map` that can bind to a version number but return some other data. I think this is problem to solve if and when it's needed though.

Issue #92, opened 2022-03-31
Textbox: add parameter for reset transform on defocus


rhelmot:
This behavior seems to differ across platforms, so we should parameterize it (and auto-detect the default based on platform?)

PR #271, opened 2022-12-20
Textbox: new interactions

### LINKED ISSUE #94: Textbox: support scroll wheel, pageup, pagedown

rhelmot:
- Scroll
- PageUp/PageDown
- Home/End

Closes #94

### DIFF
--- a/crates/vizia_core/src/context/proxy.rs
+++ b/crates/vizia_core/src/context/proxy.rs
@@ -92,12 +92,17 @@ impl ContextProxy {
     where
         F: 'static + Send + FnOnce(&mut ContextProxy),
     {
-        let mut cxp = ContextProxy {
+        let mut cxp = self.clone();
+        std::thread::spawn(move || target(&mut cxp));
+    }
+}
+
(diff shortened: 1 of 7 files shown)


geom3trik:
I've noticed that when the textbox loses focus it scrolls the content to the end even if it was previously scrolled to a different part. The same thing happens if you scroll the textbox before it has focus and then click on it to give it focus. I don't think the scroll position should reset with focus change.

rhelmot:
https://github.com/vizia/vizia/issues/92

I think this behavior is also on main? I think we had an argument about this behavior at one point.

geom3trik:
> https://github.com/vizia/vizia/issues/92
> 
> I think this behavior is also on main? I think we had an argument about this behavior at one point.

Ah okay, I'll have to see what the typical behaviour for each platform is then since you mention it varies. To me it feels wrong for the scroll to be set to the end when the focus is lost. I think either the beginning or leaving the scroll where it was makes more sense to me before knowing what the typical behaviour is. Maybe we should make it configurable?

PR #273, opened 2022-12-23
Require Eq instead of PartialEq for bindings

rhelmot:
This fixes a bug following https://github.com/vizia/vizia/pull/269 where vizia applications could take 100% cpu when any value in any model was NaN. To put it lightly, this sucks. Requiring Eq is a pain when anything is a floating point type, but this is mitigated partially by two new lenses which can be invoked to translate a float type to and from its underlying bit representation. Things get messier when you are binding to structs which contain floats - my only suggestions are to either impl PartialEq explicitly such that you can safely implement Eq, or to replace all your f32 types with a wrapper type that implements Eq, for example the eq-float crate.

If people think this is gross we can just revert #269.

### DIFF
--- a/crates/vizia_core/src/handle.rs
+++ b/crates/vizia_core/src/handle.rs
@@ -69,7 +69,7 @@ impl<'a, V> Handle<'a, V> {
     pub fn bind<L, F>(self, lens: L, closure: F) -> Self
     where
         L: Lens,
-        <L as Lens>::Target: Clone + PartialEq,
+        <L as Lens>::Target: Clone + Eq,
         F: 'static + Fn(Handle<'_, V>, L),
     {
         let entity = self.entity();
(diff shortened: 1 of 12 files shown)


Fredemus:
Good catch!

I feel like this means the ergonomics are worse compared to the old `Data` trait, at least for my use case.

geom3trik:
So thinking about this, it seems that the only way around having float values within user data structs which derive `Eq` is to new type them, versus when we had the `Data` trait and third-party types required a new-type. Between these I prefer the latter as it's much more likely for there to be float values in user-data than third-party types. Probably there will be both in a lot of cases, but needing to have special handling and special lenses just to use an `f32` seems wrong to me.

Therefore, I propose we revert #269. The effort was not wasted though, we needed to see what the alternative is to make an informed decision. So thank you!

PR #274, opened 2022-12-24
Textbox: Do not select all on start, do not scroll on end

### LINKED ISSUE #92: Textbox: add parameter for reset transform on defocus

rhelmot:
Closes #92, kind of, it doesn't actually implement the parameter described in that issue but I'm not convinced such a parameter needs to exist. I am pretty sure we have selection behavior parity with most other gui toolkits.

### DIFF
--- a/crates/vizia_core/src/views/textbox.rs
+++ b/crates/vizia_core/src/views/textbox.rs
@@ -27,7 +27,6 @@ pub struct TextboxData {
 
 impl TextboxData {
     pub fn new(text: String) -> Self {
-        let text_length = text.as_str().len();
         Self {
             text: text.clone(),
             selection: Selection::new(0, 0),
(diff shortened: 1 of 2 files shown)


crates/vizia_core/src/views/textbox.rs
```diff
@@ -861,6 +838,21 @@ where
 
             _ => {}
         });
+
+        if let Some(msg) = event.take::<TextEventInternal>() {
+            let TextEventInternal::Reset = msg;
```
geom3trik, line 843: Should this be an `if let`?

rhelmot: It would be an irrefutable if-let, so this lints better. This essentially just serves to use the variable. When the enum has more than one variant, this can be replaced with if-let.

geom3trik: Ah I see, I did wonder if it was some kind of placeholder for now.

geom3trik (approved):

PR #275, opened 2022-12-25
Replace text layout with cosmic-text and rendering with swash

### LINKED ISSUE #272: Integrate cosmic-text into Vizia
Proposal to integrate the [cosmic-text](https://github.com/pop-os/cosmic-text) crate into vizia. This would improve our current text stack and text rendering.

Currently we use [femtovg](https://github.com/femtovg/femtovg) for text rendering. Femotovg uses the RustyBuzz crate for shaping and renders glyph paths on the GPU, then stores these rendered glyphs in a font atlas, which is then used to texture quads to produce the final result.

Cosmic text provides a pipeline for font loading, font fallback, shaping (using rustybuzz), layout, and rendering (using swash), as well as text attributes (rich text) and text editing.

Since for now we want to keep femtovg for rendering everything else, I see 3 possible ways forward for integation:

1. Patch femtovg to use cosmic-text for its text pipeline. I'm not sure if this is the ideal approach as it will mean either maintaining a fork of femtovg, unless the maintainers are open to upstreaming the changes.
2. Integrate cosmic-text on top of femtovg. There's two sub-options here, (a) we can either use cosmic-text for everything *except* rendering and then render the paths with femovg (I suspect this will not look very good), or (b) we can use cosmic-text for rendering as well and maintain our own font atlas to use with femtovg (this may require at least some patching to femtovg to make this efficient).
3. Bypass femtovg for text rendering. We could go down to the OpenGL level to render text. Essentially use cosmic-text for everything but then write our own rendering code to blit the rendered glyphs. I think this is the least desiable option because it locks us in to OpenGL (femtovg uses opengl currently but could one day provide an alternative backend) and could cause problems with femtovg.

Of the options above I feel like 2(b) is the most viable at the moment, but I'm open to alternative ideas.

rhelmot:
Closes #272


rhelmot:
Status: the bridge between cosmic-text/swash and femtovg has been written. The bridge between vizia and cosmic-text has not been written.

rhelmot:
Requires my fork of femtovg (waiting on their advice in terms of how to integrate that into their main branch) and my fork of cosmic-text (https://github.com/pop-os/cosmic-text/pull/60). Outstanding issues that require more changes to cosmic-text but I'm going to put on fixing them for now:

- some bidi embedding issues
- vertical cursor movement doesn't preserve x position properly
- trying to do alignment with textboxes will fail for now
- cursor affinity

crates/vizia_baseview/src/window.rs
```diff
@@ -180,14 +175,14 @@ impl ViziaWindow {
...
-                    BackendContext::new(&mut context).text_context().resize_shaping_run_cache(size);
-                }
+                //if let Some(size) = text_shaped_words_cache_size {
+                //    BackendContext::new(&mut context)
+                //        .text_context()
+                //        .resize_shaped_words_cache(size);
+                //}
+                //if let Some(size) = text_shaping_run_cache_size {
+                //    BackendContext::new(&mut context).text_context().resize_shaping_run_cache(size);
+                //}
```
geom3trik, line 185: I suppose we can just remove this now?

crates/vizia_core/resources/themes/default_layout.css
```diff
@@ -78,6 +82,10 @@ label {
     child-left: 0px;
 }
 
+checkbox {
+    /*font-family: Entypo, sans-serif;*/
+}
```
geom3trik, line 87: Should this be uncommented?

crates/vizia_core/src/context/draw.rs
```diff
@@ -100,9 +101,9 @@ impl<'a> DrawContext<'a> {
         self.cache.get_clip_region(self.current)
     }
 
-    /// Returns the name of the default font.
-    pub fn default_font(&self) -> &str {
-        &self.style.default_font
+    /// Returns the lookup pattern to pick the default font.
+    pub fn default_font(&self) -> Vec<FamilyOwned> {
```
geom3trik, line 105: Maybe this should return by reference to avoid the clone? I checked and this isn't actually used anywhere so I don't know what the intended use would be.

crates/vizia_core/src/context/mod.rs
```diff
@@ -115,7 +130,10 @@ impl Context {
             focus_stack: Vec::new(),
             cursor_icon_locked: false,
             resource_manager: ResourceManager::new(),
-            text_context: TextContext::default(),
+            cosmic_context: CosmicContext::new_from_locale_and_db(
+                sys_locale::get_locale().unwrap_or_else(|| "en-US".to_owned()),
+                db,
+            ),
```
geom3trik, line 136: Didn't realise you need a locale to initialise the cosmic context. I guess this means we should probably try to provide it with the system locale, which you can get with the `sys_locale` crate (see `environment.rs` for example).

crates/vizia_core/src/context/mod.rs
```diff
@@ -310,6 +329,7 @@ impl Context {
             self.data.remove(*entity);
             self.views.remove(entity);
             self.entity_manager.destroy(*entity);
+            self.cosmic_context.clear_buffer(*entity);
```
geom3trik, line 332: I assume this is a no-op (more or less) if the entity doesn't have a buffer?

crates/vizia_core/src/context/mod.rs
```diff
@@ -355,19 +375,34 @@ impl Context {
...
+                new_ccx.with_buffer(entity, move |buf| {
+                    buf.lines = lines
+                        .into_iter()
+                        .map(|line| BufferLine::new(line, AttrsList::new(Attrs::new())))
+                        .collect();
+                });
+            }
+            new_ccx
+        });
     }
```
geom3trik, line 397: I guess this is required because cosmic doesn't provide a way to add fonts after the context has been created?

crates/vizia_core/src/modifiers/text.rs
```diff
@@ -1,25 +1,37 @@
...
+
+            //         cx.need_relayout();
+            //         cx.need_redraw();
+            //     }
+            // } else {
+            //     cx.style.text.insert(entity, text_data);
+
+            //     cx.need_relayout();
+            //     cx.need_redraw();
+            // }
```
geom3trik, line 34: I know it was me that commented this out but you can go ahead and delete it now.

crates/vizia_core/src/systems/layout.rs
```diff
@@ -1,24 +1,14 @@
...
-        let mut store = (Style::default(), TextContext::default(), ResourceManager::default());
-        std::mem::swap(&mut store.0, &mut cx.style);
-        std::mem::swap(&mut store.1, &mut cx.text_context);
-        std::mem::swap(&mut store.2, &mut cx.resource_manager);
-
-        layout(&mut cx.cache, &cx.tree, &store);
-        std::mem::swap(&mut store.0, &mut cx.style);
-        std::mem::swap(&mut store.1, &mut cx.text_context);
-        std::mem::swap(&mut store.2, &mut cx.resource_manager);
+        layout(&mut cx.cache, &cx.tree, &cx.style, &mut cx.cosmic_context);
```
geom3trik, line 11: Really nice to see the hack gone :D

crates/vizia_core/src/systems/text_constraints.rs
```diff
@@ -36,112 +34,62 @@ pub fn text_constraints_system(cx: &mut Context, tree: &Tree<Entity>) {
...
+                    let h = buf.layout_runs().len() as f32 * buf.metrics().line_height as f32;
+                    (w, h)
+                });
 
-                    // Add an extra pixel to account for AA
-                    let text_width = text_width.round() + 1.0 + child_space_x;
-                    let text_height = text_height.round() + 1.0 + child_space_y;
+                // Add an extra pixel to account for AA
+                let text_width = text_width.round() + 1.0 + child_space_x;
+                let text_height = text_height.round() + 1.0 + child_space_y;
```
geom3trik, line 86: So before I added the extra 1px padding for AA because of how femtovg does AA (it adds a 1px edge). I'm wondering if we need this any more?

crates/vizia_core/src/views/textbox.rs
```diff
@@ -1,40 +1,39 @@
...
 #[derive(Lens)]
 pub struct TextboxData {
-    text: String,
-    selection: Selection,
-    sel_x: f32,
-    re_sel_x: bool,
+    //text: String,
+    //selection: Selection,
+    //sel_x: f32,
+    //re_sel_x: bool,
```
geom3trik, line 16: Go ahead and remove these if we don't need them any more.

crates/vizia_core/src/views/textbox.rs
```diff
@@ -1,40 +1,39 @@
...
+        //let text_length = text.as_str().len();
         Self {
-            text: text.clone(),
-            selection: Selection::new(0, 0),
-            sel_x: -1.0,
-            re_sel_x: false,
+            //text: text.clone(),
+            //selection: Selection::new(0, text_length),
+            //sel_x: -1.0,
+            //re_sel_x: false,
```
geom3trik, line 33: Same as above.

crates/vizia_core/src/views/textbox.rs
```diff
@@ -527,52 +470,62 @@ where
...
                     td.set_caret(&mut EventContext::new(cx));
                     let parent = cx.current().parent(&cx.tree).unwrap();
                     cx.with_current(parent, |cx| td.build(cx));
                     cx.emit_to(cx.current(), ());
                 }
             });
+            let text = lens.view(cx.data().unwrap(), |text| {
+                text.map(|x| x.to_string()).unwrap_or_else(|| "".to_owned())
+            });
             TextboxContainer {}
```
geom3trik, line 514: Do we still need a container as well as the content?

crates/vizia_core/src/context/mod.rs
rhelmot: yep - this is a workaround for https://github.com/pop-os/cosmic-text/issues/57

crates/vizia_core/src/views/textbox.rs
rhelmot: I think we could _technically_ get rid of container, it would just be more computation to apply the margins between the textbox and the container. I'm fine having the layout system compute that for us.

crates/vizia_core/src/context/mod.rs
rhelmot: Is that not what's being done right now?

geom3trik: Haha, yep, I can't read. I saw the fallback to en-US and thought that was being passed in.

geom3trik (approved):

PR #277, opened 2022-12-26
Add window event to reload application stylesheets

### LINKED ISSUE #251: Add ability to reload stylesheets using an event
Currently pressing the F5 key causes stylesheets to relaod. This should be configurable and an event should be exposed so that a user can trigger a reload.

geom3trik:
Fixes #251

### DIFF
--- a/crates/vizia_window/src/window_event.rs
+++ b/crates/vizia_window/src/window_event.rs
@@ -93,4 +93,6 @@ pub enum WindowEvent {
     Relayout,
     /// Prints the debug message to the console.
     Debug(String),
+    /// Reloads all application stylesheets.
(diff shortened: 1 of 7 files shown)


crates/vizia_window/src/window_event.rs
```diff
@@ -93,4 +93,6 @@ pub enum WindowEvent {
     Relayout,
     /// Prints the debug message to the console.
     Debug(String),
+    /// Reloads all application stylesheets.
+    Reload,
```
rhelmot, line 97: Maybe name this ReloadStyles?

rhelmot (approved):

PR #282, opened 2023-01-03
Support RTL text in textboxes

### LINKED ISSUE #93: Textbox: support rtl scripts
it shouldn't be too hard, we just need to do the bidi analysis the same way femtovg does. I left TODOs in the relevant places in the code (mostly text/layout.rs)

rhelmot:
Closes #93

### DIFF
--- a/crates/vizia_core/src/views/textbox.rs
+++ b/crates/vizia_core/src/views/textbox.rs
@@ -2,7 +2,7 @@ use crate::cache::BoundingBox;
 use crate::prelude::*;
 use crate::text::{enforce_text_bounds, ensure_visible, Direction, Movement};
 use crate::views::scrollview::SCROLL_SENSITIVITY;
-use cosmic_text::{Action, Attrs, Edit, LayoutGlyph};
+use cosmic_text::{Action, Attrs, Edit};
 use std::sync::Arc;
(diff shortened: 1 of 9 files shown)


crates/vizia_core/src/views/textbox.rs
```diff
@@ -131,8 +107,12 @@ impl TextboxData {
             buf.action(match movement {
                 Movement::Grapheme(Direction::Upstream) => Action::Previous,
                 Movement::Grapheme(Direction::Downstream) => Action::Next,
+                Movement::Grapheme(Direction::Left) => Action::Left,
+                Movement::Grapheme(Direction::Right) => Action::Right,
```
geom3trik, line 111: How do these differ from upstream and downstream above? And do we  need both?

geom3trik (approved):

crates/vizia_core/src/views/textbox.rs
rhelmot: We might need both if, for example, the accessibility system has some input for "next character". We need Left and Right because they correspond to the actual intention of the user when they press the left or right arrow keys, and this corresponds to either upstream or downstream depending on whether the text is rtl or not.

geom3trik (approved):

PR #297, opened 2023-01-28
More performance improvements

geom3trik:
This PR aims to improve perforamance by targetting some 'easy wins'.

- Mouse events are now processed once per frame. If vizia receives multiple `CursorMoved` events before a `MainEventsCleared` event then the hover system is now only called once in the `MainEventsCleared`. On my system I only get one `CursorMoved` followed by one `MainEventsCleared` but it was reported that sometimes this is not always the case (possibly due to a high poll rate gaming mouse).
- All clones of the `Tree` have been removed. This required quite a bit of refactoring in some places, as well as the addition of a new kind of context, a `ResourceContext`, which is used for loading images with the `set_image_loader` method on `Context`.
- There is now a bitflag, `SystemFlags`, used for skipping systems if they aren't required to run. Before we were skipping style, layout, and draw if not needed but now we skip the other systems which run in `process_visual_updates`.
- Modifiers and the style system now selectively decide which systems need to be run. Before many of the modifiers would trigger relayout and redraw even if they didn't need to.

## TODO
- [x] Cache or find a way to remove the cloning + sorting of the tree to produce the draw tree.
    - Added a `DrawIterator` which iterates the tree in draw order, taking in to account the z-order while iterating. This removes the need for a z-order system as well as the need to clone and sort the tree by z-order. However, this now means that children can not be placed behind their parent using z-order.
- [ ] Run systems incrementally on entities. For example, when an entity has its transform changed we currently iterate the whole tree when we only need to iterate down from the changed entity. This isn't trivial though, because if multiple entities change, and some of the those are descendants of others, we need a good way to figure out the minimal set of entities that need updating and then to just update those. This becomes more complicated when we have systems such as layout and style where changing an entity could affect its ancestors (and the ancestors descendants) as well.


geom3trik:
I think I've gone about as far as I want to go on optimisation for now. Once 0.1 is released we can come back to optimisation work and focus on making everything as incremental as possible.

crates/vizia_core/src/systems/text_constraints.rs
```diff
@@ -1,27 +1,27 @@
...
-        if entity == Entity::root() {
-            continue;
-        }
-
-        if cx.cache.display.get(entity) == Some(&Display::None) {
-            continue;
-        }
-
-        if tree.is_ignored(entity) {
+        // Skip if the entity is invisible or out of bounds
```
rhelmot, line 17: this comment should probably be updated since there is no oob check

rhelmot (approved):

rhelmot:
That fix seems to have done it.

PR #298, opened 2023-01-28
Changes to allow nih-plug to depend on vizia main

geom3trik:
As far as I can tell these are the minimal set of changes required for [nih-plug](https://github.com/robbert-vdh/nih-plug) to depend on vizia main rather than a fork of vizia, which will make it easier for nih-plug to adopt changes to vizia.

- Added feature to allow disabling the default fonts from bein embedded into the binary.
- Added `set_text_config()` method on `Application` to allow configuring the hinting and subpixel rendering of text using a new `TextConfig` type.
- Updated femtovg to v0.4.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -213,13 +213,14 @@ name = "save_dialog"
 path = "examples/save_dialog.rs"
 
 [features]
-default = ["winit", "clipboard", "x11", "wayland"]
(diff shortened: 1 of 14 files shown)


rhelmot (approved):

rhelmot:
When the need for it arises in the future, we may want to add a flag to TextConfig for loading system fonts.

PR #303, opened 2023-02-14
Add tuple lens and boolean or & and lenses

geom3trik:
This PR adds three new lenses:

- The first is a tuple lens which can be used like so in a binding view:
```rust
// Rebuilds if either target data of lens1 or lens2 changes.
Binding::new(cx, (lens1, lens2), |cx, (l1, l2)|{

});
```
Or like so with a `map`:
```rust
(lens1, lens2).map(|(l1, l2)|{...});
```
This currently only works if the `Source` of lens1 is the same as the `Source` of lens2. 

This gets us one small step closer to 'derived data', which is data we can bind to that doesn't exist in a model(s) but is derived from data which does. `Map` allows us to derive data from a single lens, and now the tuple lens, combined with `map`, lets us derive data from multiple lenses of the same model (tuple lenses can be combined with other tuple lenses). Ideally in the future we would like to add what I'm calling 'multibinding', which would let us derive data from multiple lenses to different models.

- The second lens added is the `OrLens` and the third is the `AndLens`. These can be constructed from lenses which have a `Target` type of `bool` and will perform the `OR` or `AND` operation on the target values when the lens resolves. These can be used like so:
```rust
lens1.or(lens2)
lens.and(lens2)
```
Or like so:
```rust
lens1 | lens2
lens1 & lens2
```

To get that last part to work with the `|` and `&` operators I had to add a `Wrapper` type around derived lenses. This shouldn't make any practical difference to users of the crate as `Lens` is implemented for the `Wrapper`. The `BitOr` and `BitAnd` traits are then implemented for `Wrapper` lens, as well as `Then` and `Map` lenses.

The last part is maybe a little controvercial but it does allow you to do something like this:
```rust
// The HStack will be visible if either of the boolean values lensed by AppData::flag1 OR AppData::flag2 are true.
HStack::new(cx, |cx|{
    ...
})
.visibility(AppData::flag1 | AppData::flag2);
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -212,6 +212,10 @@ path = "examples/style/outline.rs"
 name = "save_dialog"
 path = "examples/save_dialog.rs"
 
+[[example]]
+name = "boolean"
+path = "examples/binding/boolean.rs"
+
 [features]
 default = ["winit", "clipboard", "x11", "wayland", "embedded_fonts"]
 clipboard = ["vizia_core/clipboard", "vizia_winit/clipboard"]
--- a/crates/vizia_core/src/environment.rs
+++ b/crates/vizia_core/src/environment.rs
@@ -1,3 +1,4 @@
+use crate::prelude::Wrapper;
 use unic_langid::LanguageIdentifier;
 use vizia_derive::Lens;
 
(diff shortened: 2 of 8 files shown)


rhelmot (approved):
No issues here. I suppose we'll see down the line whether this is for some reason incompatible with anything important :)

PR #291, opened 2023-01-17
Updated style system, built-in views, and theming

geom3trik:
# New Features

- New supported style properties (see list below).
- New supported CSS selectors (see list below).
- New and updated built-in view (see list below).
- Updated default theming including light/dark themes.
- Updated styling, including:
    - Multiple animatable box-shadows on one element.
    - Multiple animatable gradients on one element.
    - Multiple animatable transforms on one element.
- Use CSS within inline style modifiers:
```rust
Element::new(cx).border_width("5px")
```

# Fixes
- Color now supports `rgb()` function.
- Border now uses `LengthOrPercentage` for width.
- Transforms and clipping now work together correctly.

# Built-in Views

View marked with a * are new.

- [x] Button
- [x] Checkbox
- [x] Chip*
- [ ] ~~ColorPicker*~~
- [x] DatePicker*
- [x] Dropdown
- [x] Element
- [x] HStack
- [ ] Image
- [ ] Knob
- [x] Label
- [ ] List
- [ ] Menu
- [x] Notification*
- [x] PickList*
- [ ] Popup
- [x] RadioButton
- [ ] Scrollbar
- [ ] ScrollView
- [x] Slider
- [x] Spinbox*
- [x] Switch*
- [ ] TabView*
- [ ] ~~Table~~
- [x] Textbox
- [ ] TimePicker*
- [ ] Tooltip*
- [x] ~~XYPad*~~

# Supported Style Properties

<details>
<summary>List of properties</summary>

## Display

- [x] display
- [x] visibility
- [x] opacity

## Clipping

- [x] overflow
- [x] clip-path (inset only)

## Filtering
- [x] backdrop-filter
- [ ] filter 

## Layout

- [x] layout-type
- [x] position-type

- [x] space
  - [x] left
  - [x] right
  - [x] top
  - [x] bottom

- [x] min-space
  - [x] min-left
  - [x] min-right
  - [x] min-top
  - [x] min-bottom

- [x] max-space
  - [x] max-left
  - [x] max-right
  - [x] max-top
  - [x] max-bottom  

- [x] size 
  - [x] width
  - [x] height

- [x] min-size
  - [x] min-width
  - [x] min-height

- [x] max-size
  - [x] max-width
  - [x] max-height

- [x] child-space
  - [x] child-left
  - [x] child-right
  - [x] child-top
  - [x] child-bottom
- [x] row-between
- [x] col-between

## Border

- [x] border
  - [x] border-color
  - [x] border-width

- [x] border-corner-shape
[...]

PR #307, opened 2023-02-14
Border Width Specialization

DasLixou:
This has currently some `TODO:`'s where i dont know any further and need help from the community 😅

### DIFF
--- a/crates/vizia_core/src/animation/anim_ext.rs
+++ b/crates/vizia_core/src/animation/anim_ext.rs
@@ -59,7 +59,10 @@ pub trait AnimExt: Copy + Sized {
 
         // Border
         cx.style.border_color.play_animation(self.entity(), animation);
-        cx.style.border_width.play_animation(self.entity(), animation);
+        cx.style.border_width_left.play_animation(self.entity(), animation);
+        cx.style.border_width_right.play_animation(self.entity(), animation);
+        cx.style.border_width_top.play_animation(self.entity(), animation);
+        cx.style.border_width_bottom.play_animation(self.entity(), animation);
         cx.style.border_radius_bottom_left.play_animation(self.entity(), animation);
         cx.style.border_radius_top_left.play_animation(self.entity(), animation);
         cx.style.border_radius_bottom_right.play_animation(self.entity(), animation);
--- a/crates/vizia_core/src/context/draw.rs
+++ b/crates/vizia_core/src/context/draw.rs
@@ -123,7 +123,10 @@ impl<'a> DrawContext<'a> {
         physical * self.style.dpi_factor as f32
     }
 
-    style_getter_units!(border_width);
+    style_getter_units!(border_width_left);
+    style_getter_units!(border_width_right);
+    style_getter_units!(border_width_top);
+    style_getter_units!(border_width_bottom);
     style_getter_units!(border_radius_top_right);
     style_getter_units!(border_radius_top_left);
     style_getter_units!(border_radius_bottom_right);
--- a/crates/vizia_core/src/layout/node.rs
+++ b/crates/vizia_core/src/layout/node.rs
@@ -230,28 +230,28 @@ impl<'w> Node<'w> for Entity {
     }
 
     fn border_left(&self, store: &Self::Data) -> Option<morphorm::Units> {
-        store.border_width.get(*self).cloned().map(|v| match v {
+        store.border_width_left.get(*self).cloned().map(|v| match v {
             Units::Pixels(val) => Units::Pixels(val * store.dpi_factor as f32),
             t => t,
         })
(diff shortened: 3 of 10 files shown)


geom3trik:
There is a bit of a blocker on this (and the main reason it's not been added before), which is that the crate we currently use for rendering (femtovg) does not support variable stroke widths. It is possible however that something like this might be achievable, perhaps by stroking each side individually (this would need to take into account border-radius and border-corner-shape).

DasLixou:
Ye I also stumbled across this issue. :c

DasLixou:
I also came up with a solution for the problem of setting everything manually or for all. You had a todo for corner radius for that too. Should I try this in this PR or make another one from main without these changes here?

DasLixou:
> > I also came up with a solution for the problem of setting everything manually or for all. You had a todo for corner radius for that too. Should I try this in this PR or make another one from main without these changes here?
> 
> I'm sorry I'm not sure I understand what you're referring to. Could you elaborate?
> 
> Also note that styling in vizia is currently being overhauled, the progress for which can be found in #291. It might be worth looking at this and targeting that branch rather than main.

Ok this PR looks quite overwhelming and i'm not sure if it contains this already (i only found Rect but it seems more like a parser rect). My Idea was just a simple `QuadUnit` struct which can be made from tuples such as `(Unit)` `(Unit, Unit)` or `(Unit, Unit, Unit, Unit)` and then get converted into `x`, `y`, `width`, `height` values and then on a element i would be able to do:
```rust
Button::new(cx, /* ... */).border_radius((Pixels(20.), Pixels(12.)));
```

DasLixou:
Hmm.. just visited this again and came up with an idea.. what about splitting the border into 4 paths? I don't know how this would look exactly but we could do some vector thing to see where the stroke is "looking" into and therefor set if its top, left, right or bottom. with that we could potentially also make circles with custom 4 side border and let them look like in web

geom3trik:
> Hmm.. just visited this again and came up with an idea.. what about splitting the border into 4 paths? I don't know how this would look exactly but we could do some vector thing to see where the stroke is "looking" into and therefor set if its top, left, right or bottom. with that we could potentially also make circles with custom 4 side border and let them look like in web

In principle you could do this fairly easily. The current drawing code actually already draws the shape using separate beziers if the order-radius isn't the same on each corner. However, I'm not sure this would look correct for drawing separate borders as the paths can only have a single thickness. If we look at what web does when there's an individual border mixed with border-radius:
image
Note that the border tapers at the top and bottom to blend with the 0px borders at the top and bottom. 

I think maybe the best way to handle this is to draw the border as a filled shape using beziers, rather than as lines with a stroke.

DasLixou:
> > Hmm.. just visited this again and came up with an idea.. what about splitting the border into 4 paths? I don't know how this would look exactly but we could do some vector thing to see where the stroke is "looking" into and therefor set if its top, left, right or bottom. with that we could potentially also make circles with custom 4 side border and let them look like in web
> 
> In principle you could do this fairly easily. The current drawing code actually already draws the shape using separate beziers if the order-radius isn't the same on each corner. However, I'm not sure this would look correct for drawing separate borders as the paths can only have a single thickness. If we look at what web does when there's an individual border mixed with border-radius: image Note that the border tapers at the top and bottom to blend with the 0px borders at the top and bottom.
> 
> I think maybe the best way to handle this is to draw the border as a filled shape using beziers, rather than as lines with a stroke.

Oh yeah i like the idea of bezier curves! :D But we want left, right etc for the border thickness, for the corner radius i think you've splitted the border into top-left, top-right, bottom-left, etc. because you need the corners.

geom3trik:
> Oh yeah i like the idea of bezier curves! :D But we want left, right etc for the border thickness, for the corner radius i think you've splitted the border into top-left, top-right, bottom-left, etc. because you need the corners.

Yes border-radius is already handled for the different corners. I was just showing what affect it has on the border edges. Something similar happens when the border widths are different:
image
See how it blends from one width to the other. Made more noticeable with the border-radius applied.

DasLixou:
> See how it blends from one width to the other. Made more noticeable with the border-radius applied.

yes that blends sooo smooth :o 

but how hard is this to implement?

DasLixou:
just came back to this and also prototyped this on the web but here i dont get those curves.
image
what did you do?

geom3trik:
> just came back to this and also prototyped this on the web but here i dont get those curves.
> image
> what did you do?

I just added a border to two sides of a rounded rectangle. What did you do?

DasLixou:
> I just added a border to two sides of a rounded rectangle. What did you do?

Just a div with exact width and height, border-radius, border-style: solid, and border-width: 50px 0px 0px 0px

PR #308, opened 2023-02-14
Implement `is_empty`

DasLixou:
Ok thats a small line of code but i think that should do it. I didn't found any tests so on your risk 😅
Also: from lines like https://github.com/vizia/vizia/blob/new-style-system/crates/vizia_style/selectors/matching.rs#L993 i think selectors like `:first-child` and `:last-child` should be also done but ill check on this again.

### DIFF
--- a/crates/vizia_core/src/systems/style.rs
+++ b/crates/vizia_core/src/systems/style.rs
@@ -78,7 +78,7 @@ impl<'s, 't, 'v> Element for Node<'s, 't, 'v> {
     }
 
     fn is_empty(&self) -> bool {
-        self.tree.has_children(self.entity)
+        !self.tree.has_children(self.entity)
     }
 
     fn is_root(&self) -> bool {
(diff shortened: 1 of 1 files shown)


DasLixou:
https://github.com/vizia/vizia/blob/new-style-system/crates/vizia_style/selectors/matching.rs#L800
Yes seems like after this PR all `Tree-structural pseudo-classes` will be done 🤩

geom3trik:
I know there aren't any unit tests for these but did you test them in your own example? I'm curious if the Nth child ones just work as well.

DasLixou:
> I know there aren't any unit tests for these but did you test them in your own example? I'm curious if the Nth child ones just work as well.

So `is_empty` works! but the nth-child ones are messed up.. going to fix those

DasLixou:
YES! everything works now 🥳🎉🪅
The problem with nth-child was just that i had the css on the parent (`.parent:last-child`) and not on the child (`.subeles:last-child`).
This is now ready to get merged and mark every Tree-structural pseudo-class as done :D

PR #309, opened 2023-02-14
Svg Displaying

DasLixou:
image

 Before this gets merged, theres some ToDo's:

* [x] Setting all params for stroke
* [x] Override Color
* [ ] ~Make the whole `impl Res<SvgTree>` thing and especially `in self.tree.result.clone()` more performant and simpler.~ currently not possible
* [x] Make the Svg in the Size of the `SvgDisplay` Component (how do i scale a path? idk)


crates/vizia_core/src/views/svg_display.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    fn event(&mut self, _cx: &mut EventContext, event: &mut Event) {
+        if let Some(event) = event.take::<SvgEvent>() {
+            match event {
+                SvgEvent::TreeChanged(new_tree) => {
+                    self.tree.replace(new_tree);
+                }
+            }
+        }
```
sky1e, line 76: This could be
```suggestion
        if let Some(SvgEvent::TreeChanged(new_tree)) = event.take::<SvgEvent>() {
            self.tree.replace(new_tree);
        }
```

DasLixou:
Thanks for the review, I will take note of that but currently this pr is blocked for #312

PR #310, opened 2023-02-15
Allow camelcase warning on struct root in lens.rs

adriannic:


geom3trik (approved):
Looks good, thanks for the fix!

PR #311, opened 2023-02-15
Color rework

DasLixou:
This makes the old Color functions constants and adds the simple Material Colors from https://materialui.co/colors/ (simple = except A***)


geom3trik:
This is not really the way I'd like to go with vizia for now, but I think material colors could be added through a trait as a separate crate on top of vizia.

PR #312, opened 2023-02-15
Add `LensValue` enum to avoid cloning lensed data

geom3trik:


crates/vizia_core/src/state/lens.rs
```diff
@@ -195,8 +220,11 @@ where
     type Source = A::Source;
     type Target = B::Target;
 
-    fn view<O, F: FnOnce(Option<&Self::Target>) -> O>(&self, source: &Self::Source, map: F) -> O {
-        self.a.view(source, |t| if let Some(t) = t { self.b.view(t, map) } else { map(None) })
+    fn view<'a>(&self, source: &'a Self::Source) -> Option<LensValue<'_, 'a, Self::Target>> {
+        match self.a.view(source) {
+            Some(LensValue::Borrowed(val)) => self.b.view(val),
```
sky1e, line 225:
```suggestion
        match self.a.view(source)? {
            LensValue::Borrowed(val) => self.b.view(val),
```

crates/vizia_core/src/state/lens.rs
```diff
@@ -312,8 +333,8 @@ impl<T: 'static> Lens for UnwrapLens<T> {
     type Source = Option<T>;
     type Target = T;
 
-    fn view<O, F: FnOnce(Option<&Self::Target>) -> O>(&self, source: &Self::Source, map: F) -> O {
-        map(source.as_ref())
+    fn view<'a>(&self, source: &'a Self::Source) -> Option<LensValue<'_, 'a, Self::Target>> {
+        source.as_ref().map(|t| LensValue::Borrowed(t))
```
sky1e, line 337:
```suggestion
        source.as_ref().map(LensValue::Borrowed)
```

crates/vizia_core/src/state/lens.rs
```diff
@@ -367,17 +387,17 @@ where
...
-                map(Some(&(num / den)))
+                Some(LensValue::Owned(num / den))
             } else {
-                map(None)
+                None
             }
         } else {
-            map(None)
+            None
         }
```
sky1e, line 401:
```suggestion
        let num = self.numerator.view(source)?.into_owned();
        let den = self.denominator.view(source)?.into_owned();
        Some(LensValue::Owned(num / den))
```

sky1e (commented):
Mostly a bunch of small clarity suggestions. i verified that everything compiles without warnings and tests pass with these changes.

My first time using GitHub diff suggestions. Hopefully I used them correctly.

crates/vizia_derive/src/lens.rs
```diff
@@ -342,12 +342,13 @@ fn derive_enum(input: &syn::DeriveInput) -> Result<proc_macro2::TokenStream, syn
                 type Source = #enum_type;
                 type Target = #variant_type;
 
-                fn view<O, F: FnOnce(Option<&Self::Target>) -> O>(&self, source: &#enum_type, map: F) -> O {
-                    map(if let #enum_type::#variant_name(inner_value) = source {
-                        Some(inner_value)
+                fn view<'a>(&self, source: &'a #enum_type) -> Option<LensValue<<'a, Self::Target>> {
```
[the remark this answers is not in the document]
geom3trik: You're totally right and no I don't think this is used or tested in vizia.

geom3trik:
> Mostly a bunch of small clarity suggestions (because @rhelmot asked me to look at the PR). i verified that everything compiles without warnings and tests pass with these changes.
> 
> My first time using GitHub diff suggestions. Hopefully I used them correctly.

Thanks for the suggestions. I didn't realise gh had that feature but after some trial and error with the gh UI I've managed to apply all your suggestions.

crates/vizia_core/src/state/lens.rs
```diff
@@ -18,13 +19,64 @@ pub trait Lens: 'static + Clone {
...
+    /// A reference to model data
+    Borrowed(&'b T),
+    /// Owned data
+    Owned(T),
+}
+
+impl<T: Clone> Clone for LensValue<'_, '_, T> {
+    fn clone(&self) -> Self {
+        match self {
+            LensValue::Borrowed(v) => LensValue::Owned((*v).clone()),
```
sky1e, line 42: Could you just return a Borrowed here instead of cloning? I don't see what that accomplishes.
```suggestion
            LensValue::Borrowed(v) => LensValue::Borrowed(*v),
```

geom3trik: Yes, good point. I'm not sure what I was thinking with that.

crates/vizia_core/src/state/lens.rs
```diff
@@ -195,8 +220,11 @@ where
     type Source = A::Source;
     type Target = B::Target;
 
-    fn view<O, F: FnOnce(Option<&Self::Target>) -> O>(&self, source: &Self::Source, map: F) -> O {
-        self.a.view(source, |t| if let Some(t) = t { self.b.view(t, map) } else { map(None) })
+    fn view<'a>(&self, source: &'a Self::Source) -> Option<LensValue<'_, 'a, Self::Target>> {
+        match self.a.view(source)? {
+            LensValue::Borrowed(val) => self.b.view(val),
+            _ => None,
+        }
```
sky1e, line 227: There are more cases that are possible to handle here.
```suggestion
        let val = self.a.view(source)?;
        let val = match val {
            LensValue::Borrowed(val) => return self.b.view(val),
            LensValue::Local(val) => val,
            LensValue::Owned(ref val) => &val,
        };
        match self.b.view(&val)? {
            LensValue::Local(val) => Some(LensValue::Local(val)),
            LensValue::Owned(val) => Some(LensValue::Owned(val)),
            _ => None,
        }
```

DasLixou: Why can’t we just store everything in Val and then do anything also for Borrow? Why these 2 extra rules for Borrow? (The direct calling and the exclude from the second match)

geom3trik: I can't check right now but I assume it's because the lifetimes are different so it would be returning different types from the match, which isn't possible. I can double check that later though.

PR #313, opened 2023-02-15
Make the impossible possible

DasLixou:
Just say that the `Target` of `B` should be `Clone`able, this is nothing new, even `LensExt`' get method does that.


geom3trik:
As discussed on Discord, making the target `Clone` is not viable because intermediate lensed data when chaining `Then` lenses may not be `Clone`. Only the last in the chain which is bound to must implement `Clone`. Therefore, I'm going to close this PR for now.

PR #314, opened 2023-02-17
Implement `Data` for all `PartialEq`'s

DasLixou:


geom3trik:
We tried this before (see #269 and #273) but there's an issue that for floating point types there are different kinds of `NaN`, which can cause bindings to constantly update every frame. The idea behind the `Data` trait is that comparisons need to be fast, which is not always true for `PartialEq`. For example, for floating point types we just do a bitwise comparison.

PR #315, opened 2023-02-17
Optimize vizia_core::state::binding::Binding::remove() to not iterate…

sky1e:
… through entire store on every remove. Additionally, fix or silence clippy warnings through the rest of binding.rs

I was profiling [Arborio](https://github.com/rhelmot/arborio)'s opening and closing maps and noticed a rather significant portion of time was spent removing children from bindings. Looking into the remove function, the entire model data store hashmap is iterated over when only a single element in it was mutated, which we still have a reference to. This change checks only that single store, and if appropriate, removes it.

This brought closing maps in Arborio down from sometimes almost half a second to basically instantaneous.

Additionally since I was looking at the file, I addressed all of the clippy warnings in it. The explicit lifetime in BindingHandler::update was pointless, so I removed it. The new() method not returning Self makes sense within this context, and I don't think a type alias for Binding::content would improve readability, so I silenced those warnings.

### DIFF
--- a/crates/vizia_core/src/state/binding.rs
+++ b/crates/vizia_core/src/state/binding.rs
@@ -13,6 +13,7 @@ where
 {
     entity: Entity,
     lens: L,
+    #[allow(clippy::type_complexity)]
     content: Option<Box<dyn Fn(&mut Context, L)>>,
 }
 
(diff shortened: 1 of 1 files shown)


geom3trik (approved):
Good spot! Thanks for the optimisation and for addressing some of the clippy warnings. I'm glad to see this makes such a big difference for arborio.

PR #316, opened 2023-02-17
Add pub method to call sync_styles on current entity's text

rhelmot:
This is necessary if you want to have a custom draw method that uses the text draw stuff, lol

### DIFF
--- a/crates/vizia_core/src/context/draw.rs
+++ b/crates/vizia_core/src/context/draw.rs
@@ -164,6 +164,10 @@ impl<'a> DrawContext<'a> {
         self.cache.get_opacity(self.current)
     }
 
+    pub fn sync_text_styles(&mut self) {
+        self.text_context.sync_styles(self.current, &self.style);
+    }
+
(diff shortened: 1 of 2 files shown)


geom3trik:
Is this not an implementation detail though? Should we not expose some method to render text ina custom draw and call this internally? Or I guess is it not so simple because you need entities for the buffers?

rhelmot:
uhhhhhhhhh I suppose this could be done automatically before drawing but we would need to somehow cache and then invalidate the synced status. not sure how to do that.

geom3trik:
hmmm right. I'm happy to merge this solution for now and we can come back to it if we can think of a beter solution later. Do you have an example of a custom draw function with text for how to use this? Presumably you just call it before calling `draw_text`?

rhelmot:
Yes - you need to call it before calling any of draw_text, draw_highlight, or draw_caret, and the results are valid until the text or any of its properties are updated. My main reason for not wanting to do a cache yet upon some further reasoning is that this will interact weirdly with text spans, and I would rather fast-track spans.

PR #317, opened 2023-02-18
Run cargo clippy --fix --all

sky1e:
I'd like to work on fixing clippy lints, and I thought I'd start by isolating automatic fixes into their own pull request for easier review. This is just the result of running `cargo clippy --fix --all` and then a manual review to check if any of the changes looked questionable.


geom3trik (approved):
These changes look good to me. Thank you for taking on the task of fixing the clippy warnings. I've tried and failed to fix these 2 times before. I've tested these changes with a few of the examples and they seem to work as expected.

PR #318, opened 2023-02-19
Fix parsing short color codes with alpha

sky1e:
Previously, when parsing a short color code with alpha, it read byte 0 into the alpha, which is actually the red channel. This fixes that, as well as refactors the surrounding code for clarity and adds a test.

### DIFF
--- a/crates/vizia_core/src/style/color.rs
+++ b/crates/vizia_core/src/style/color.rs
@@ -103,18 +103,16 @@ impl From<&str> for Color {
         match clean_hex.len() {
             3 | 4 => {
                 let hex = clean_hex.as_bytes();
-                let r = (hex[0] as char).to_digit(16).unwrap() as u8 * 17;
-                let g = (hex[1] as char).to_digit(16).unwrap() as u8 * 17;
-                let b = (hex[2] as char).to_digit(16).unwrap() as u8 * 17;
-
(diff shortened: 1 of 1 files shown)


DasLixou (approved):

geom3trik (approved):

DasLixou (commented):
Why do we need to multiply a `u8` by `0x11`?

geom3trik:
> Why do we need to multiply a `u8` by `0x11`?

It converts the short form single hex digit, e.g. 4 into the two digit hex 44. It's equivalent to multiplying by 11 in decimal.

sky1e:
> Why do we need to multiply a `u8` by `0x11`?

It's to turn the range 0x0..=0xF into 0x00..=0xFF, turning one hex digit into two. It corresponds to the multiply by 17 in the replaced lines (0x11 == 17).

geom3trik:
> > Why do we need to multiply a `u8` by `0x11`?
> 
> It's to turn the range 0x0..=0xF into 0x00..=0xFF, turning one hex digit into two. It corresponds to the multiply by 17 in the replaced lines (0x11 == 17).

Ah yeh sorry, I meant that it does the same for hex as multiplying by 11 does for decimal. Not that 0x11 == 11. Probably not a useful comparison 😅

DasLixou:
Oh ok makes sense thanks 😊

PR #323, opened 2023-02-27
Adding `relative_position`

DasLixou:
* Adding `relative_position` for `EventContext` returning `RelativeResult`
* Adding `bounds` helper for `EventContext`

### DIFF
--- a/crates/vizia_core/src/context/event.rs
+++ b/crates/vizia_core/src/context/event.rs
@@ -27,6 +27,17 @@ pub enum RelativeResult {
     Outside((f32, f32)),
 }
 
+impl RelativeResult {
+    /// Returns the `x` and `y` coordinates. May be negative when unwrapping [`RelativeResult::Outside`]
+    #[inline]
+    pub fn unwrap(self) -> (f32, f32) {
+        match self {
+            Self::Inside((x, y)) => (x, y),
+            Self::Outside((x, y)) => (x, y),
+        }
+    }
+}
+
 pub struct EventContext<'a> {
     pub(crate) current: Entity,
     pub(crate) captured: &'a mut Entity,
--- a/crates/vizia_core/src/context/event.rs
+++ b/crates/vizia_core/src/context/event.rs
@@ -5,7 +5,7 @@ use std::error::Error;
 
 use fnv::FnvHashMap;
 
-use crate::cache::CachedData;
+use crate::cache::{BoundingBox, CachedData};
 use crate::events::ViewHandler;
 use crate::prelude::*;
 use crate::resource::ResourceManager;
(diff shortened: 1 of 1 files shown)


DasLixou:
I am kinda unsure about the Inside / Outside thing because you will mostly use this with mouse events and they are as far as I know only sent if its "inside" the view, aren't they?

geom3trik:
> I am kinda unsure about the Inside / Outside thing because you will mostly use this with mouse events and they are as far as I know only sent if its "inside" the view, aren't they?

Mouse events can be sent to the view when outside of its bounds if the view has captured mouse events. This is typically done when a view is draggable so that the `MouseUp` event is still received if the mouse leaves the view while dragging.

DasLixou:
> > I am kinda unsure about the Inside / Outside thing because you will mostly use this with mouse events and they are as far as I know only sent if its "inside" the view, aren't they?
> 
> Mouse events can be sent to the view when outside of its bounds if the view has captured mouse events. This is typically done when a view is draggable so that the `MouseUp` event is still received if the mouse leaves the view while dragging.

Thats my thought. So what do you say is better? Leave it as it is or make it just directly return a `(f32, f32)`. But you can now just do `relative_position(x, y).unwrap()` anyways so i think it wouldn't be bad to have that extra thing, right?

geom3trik:
> Thats my thought. So what do you say is better? Leave it as it is or make it just directly return a (f32, f32). But you can now just do relative_position(x, y).unwrap() anyways so I think it wouldn't be bad to have that extra thing, right?

I like how you had it with the enum. But it depends on how we expect this to be used. Essentially we're encoding 2 pieces of information into one here, the relative position and also whether it's inside/outside. Does it become inconvenient if all you need is one or the other? Another potential solution is to have a separate method for checking if the relative position is inside the bounds of the view. I think `BoundingBox` may already have this method. Let's see if anyone else has any opinions on this.

DasLixou:
yeah but boundingbox can only check for containing other boundingboxes. and this seems pretty clean to me (here are my two usages):
```rust
WindowEvent::MouseDown(_) => {
    if let RelativeResult::Inside((rel_x, rel_y)) =
        cx.relative_position(cx.mouse.cursorx, cx.mouse.cursory)
    {
        // implementation
    }
}

WindowEvent::MouseMove(x, y) => {
    let (rel_x, rel_y) = cx.relative_position(*x, *y).unwrap();
    // implementation
}
```

DasLixou:
But for .unwrap() I force the user to a for them maybe unused isInside check

geom3trik:
I think for this we could rename it to something like `cursor_pos_relative()`, and rather than passing an `x` and `y` to the function instead just return the relative position for the mouse cursor. Because I imagine this is the primary use-case for this. Also I'm not a fan of `unwrap` here, could we come up with a different name?

PR #326, opened 2023-02-28
Add example for custom images

geom3trik:


DasLixou:
Why not create the image in the constructor? Would get rid of the option

geom3trik:
> Why not create the image in the constructor? Would get rid of the option

Because you need the `canvas` supplied by the `draw` function to create the image id. This example is a workaround for an issue where images loaded with vizia aren't rendering with the baseview backend. Ideally we want vizias resource manager to manage images rather than a custom view like this, but I thought I'd make this example so people aren't blocked while I'm finding the source of the bug.

DasLixou:
ok

PR #328, opened 2023-03-05
Pass lenses by reference to avoid cloning into closures

geom3trik:


crates/vizia_core/src/views/checkbox.rs
```diff
@@ -133,7 +133,7 @@ impl Checkbox {
     /// #
     /// Checkbox::new(cx, AppData::value);
     /// ```
-    pub fn new(cx: &mut Context, checked: impl Lens<Target = bool>) -> Handle<Self> {
+    pub fn new<'a>(cx: &'a mut Context, checked: &impl Lens<Target = bool>) -> Handle<'a, Self> {
```
rhelmot, line 136: fyi: this is not necessary, lifetime elision can handle it, whatever linter is yelling at you about it is lying to you. clippy will yell about this form.

geom3trik:
Closing in favour of #342

PR #342, opened 2023-06-13
Make `Lens` `Copy`

geom3trik:
Adds a thread local store for lens mapping function, allowing the `Map` lens to be made `Copy`. This in turn allows `Lens` itself to be `Copy`.

In current vizia derived lenses are `Copy` but the `Lens` trait itself does not have a `Copy` bound because `Map` lenses are not `Copy`. This is because current `Map` lenses store a closure (for the mapping from lens output to desired output). This causes problems when passing lenses into closures and usually requires creating a clone of the lens to move into a closure.

This change solves this problem by moving the mapping closure out of the `Map` lens and into a thread local. Now the `Map` contains an id into a store of mapping functions. I'm not typically a fan of using global data like this but the ergonomic improvements are, I believe, worth it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -325,6 +325,10 @@ path = "examples/custom_modifiers.rs"
 name = "keyframes"
 path = "examples/keyframes.rs"
 
+[[example]]
+name = "lens_map"
+path = "examples/lens_map.rs"
+
 
 [features]
 default = ["winit", "clipboard", "x11", "wayland", "embedded_fonts"]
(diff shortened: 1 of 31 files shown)


geom3trik:
If this change is deemed appropriate at this stage then I'll add to this PR later by removing the unnecessary lens cloning, which will fix the clippy warnings and thus appease the CI gods.

geom3trik:
I've tested this quite a bit now and I'm happy that it's working as intended. I also fixed a bug that was causing a memory leak where stores made from maps would never be deleted because the call to `cashe_key` would generate a new id every time. For these reasons I'm going to go ahead and merge this now.

PR #353, opened 2023-06-25
Allow `Datepicker` to be localized

geom3trik:
Adds `ToStringLocalized`, which allows lenses which target `Localized`. This allows the months and days used by the datepicker to be localized.

To localize the datepicker, simply add the appropriate key/values to an ftl file. See `examples/resources/translations/fr/hello.ftl` for example.

### DIFF
--- a/crates/vizia_core/src/binding/data.rs
+++ b/crates/vizia_core/src/binding/data.rs
@@ -68,6 +68,7 @@ impl_data_simple!(NaiveTime);
 impl_data_simple!(Angle);
 impl_data_simple!(String);
 impl_data_simple!(Entity);
+impl_data_simple!(Localized);
 
 impl Data for &'static str {
     fn same(&self, other: &Self) -> bool {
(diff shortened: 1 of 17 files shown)


geom3trik:
Fixes #353

iron3oxide:
Great! I'm admittedly not that great at understanding other peoples Rust code yet, but sifting through the diff view it seemed t me that this enables other widgets/views to be localized as well, doesn't it?

geom3trik:
> Great! I'm admittedly not that great at understanding other peoples Rust code yet, but sifting through the diff view it seemed to me that this enables other widgets/views to be localized as well, doesn't it?

Yep that's right. It was already possible before to pass a `Localized` to views in order to have localized (translated) text, but it's now possible to pass a lens to a `Localized`, which was necessary for the datepicker which internally used a picklist which expects a lens to a vec of data. Now this data can be a vec of `Localized`, which is now what it does. In case you wanted to know the details.

I think the datepicker was the only view using English text but let me know if you find anything else. Ty.

iron3oxide:
That's good to know, always helpful to have some details. Haven't found anything else with English text either, glad that opening the issue was of help. Impressive time to fix btw!

PR #354, opened 2023-06-27
Timed events

geom3trik:


geom3trik:
Supersedes #24

crates/vizia_core/src/context/backend.rs
```diff
@@ -264,4 +265,19 @@ impl<'a> BackendContext<'a> {
     pub fn needs_refresh(&mut self) {
         self.0.style.system_flags = SystemFlags::all();
     }
+
+    pub fn emit_scheduled_events(&mut self) {
```
DasLixou, line 269: Shouldnt this be called like "wait_for_scheduled_events"?

DasLixou:
So when I understand this correctly we are constantly popping and pushing in the BinaryHeap until a finished scheduled event is placed at the top, then we get this and execute it, right? Wouldn't it be smarter to just sort them from their time once and maybe use something like wakers in the async world instead of constantly whiling over it?

geom3trik:
> So when I understand this correctly we are constantly popping and pushing in the BinaryHeap until a finished scheduled event is placed at the top, then we get this and execute it, right? Wouldn't it be smarter to just sort them from their time once and maybe use something like wakers in the async world instead of constantly whiling over it?

No that's not what's happening. A binary heap is already sorted, so each cycle it checks which scheduled events are ready to be sent and then pushes those events into the event queue by popping them off of the binary heap. These are two different structures, `event_queue` and `event_schedule`.

crates/vizia_core/src/context/backend.rs
geom3trik: Sure but you're also emitting those events by moving them to the event queue.

PR #356, opened 2023-06-30
Update readme

geom3trik:
- Add more comprehensive list of features.
- Add link to book.
- Update counter image.


iron3oxide:
Looks nice! An idea that just came to me: why not add a small section regarding icons? Maybe mention that 4250+ icons are theoretically available, how one can add a not yet added icon and link to tabler.io for all the available ones

geom3trik:
> Looks nice! An idea that just came to me: why not add a small section regarding icons? Maybe mention that 4250+ icons are theoretically available, how one can add a not yet added icon and link to tabler.io for all the available ones

That's a good idea, though I wonder if that's something that should be in the readme or just have a dedicated page in the book? Maybe both?

iron3oxide:
Maybe one could mention "4250+ icons provided by tabler" as a feature and add the details in a book page

PR #372, opened 2023-07-12
added the circle drawer example for 7gui

YouKnow-sys:
added the circle drawer example for 7gui
I don't like how the undo/redo works atm that much, so I'm open to suggestions to improve it.
any other suggestion for improvement is more then welcome.

https://github.com/vizia/vizia/assets/17937017/29d5602a-4664-4f82-a290-19eba4b57dd2

### DIFF
--- a/examples/7GUIs/circle_drawer.rs
+++ b/examples/7GUIs/circle_drawer.rs
@@ -59,17 +59,92 @@ struct CircleData {
     selected: Option<usize>,
 }
 
+impl CircleData {
+    fn add_circle(&mut self, circle: Circle) {
+        self.selected = Some(self.circles.len());
+        self.circles.push(circle);
+    }
+
+    fn change_radius(&mut self, r: f32) {
+        if let Some(idx) = self.selected {
+            self.circles[idx].r = r;
+        }
+    }
+
+    fn update_selected(&mut self, x: f32, y: f32) {
+        self.selected = self
+            .circles
+            .iter()
+            .enumerate()
+            .rev()
+            .find(|(_, c)| distance(c.x, c.y, x, y) < c.r)
+            .map(|(i, _)| i);
+    }
+
+    fn get_selected_radius(&self) -> Option<f32> {
+        self.selected.map(|idx| self.circles[idx].r)
+    }
+}
+
 enum UndoRedoAction {
     Circle(Circle),
     RadiusChange(usize, f32),
 }
 
-#[derive(Lens)]
+#[derive(Default, Lens)]
+struct UndoRedo {
+    undo_list: Vec<UndoRedoAction>,
+    redo_list: Vec<UndoRedoAction>,
+}
+
+impl UndoRedo {
+    fn add_action(&mut self, action: UndoRedoAction) {
+        self.undo_list.push(action);
+        self.redo_list.clear(); // empty the redo list
+    }
+
+    fn undo(&mut self, circles: &mut Vec<Circle>) {
+        let last = self.undo_list.pop().unwrap();
+
+        match last {
+            UndoRedoAction::Circle(_) => {
+                self.redo_list.push(last);
+                circles.pop(); // remove the last circle
+            }
+            UndoRedoAction::RadiusChange(idx, r) => {
+                self.redo_list.push(UndoRedoAction::RadiusChange(
+                    idx,
+                    circles[idx].r, // store the current radius in redo list
+                ));
+                circles[idx].r = r; // update the radius to the old one
+            }
+        }
+    }
+
+    fn redo(&mut self, circles: &mut Vec<Circle>) {
+        let last = self.redo_list.pop().unwrap();
+
+        match last {
+            UndoRedoAction::Circle(c) => {
+                self.undo_list.push(last);
+                circles.push(c);
+            }
+            UndoRedoAction::RadiusChange(idx, r) => {
+                self.undo_list.push(UndoRedoAction::RadiusChange(
+                    idx,
+                    circles[idx].r, // store the current radius in undo list
+                ));
+                circles[idx].r = r; // restore the radius
+            }
+        }
+    }
+}
+
+#[derive(Default, Lens)]
 struct CircleDrawerData {
     circles_data: CircleData,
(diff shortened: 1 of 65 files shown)


examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    .dialog-box {
+        child-space: 1s;
+        width: 460px;
+        height: 100px;
+        top: 1s;
+        left: 1s;
+        right: 1s;
+        bottom: 40px;
```
geom3trik, line 31: This can be shortened to:
```css
space: 1s;
bottom: 40px;
```

examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+        background-color: rgba(255, 255, 255, 0.7);
+        border-color: black;
+        border-width: 1px;
```
geom3trik, line 34: This can be shortened to:
```css
border: 1px black;
```

examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+#[derive(Clone, Copy, Data)]
+struct Circle {
+    x: f32,
+    y: f32,
+    r: f32,
+}
+
+#[derive(Clone, Default, Data, Lens)]
+struct CircleData {
+    /// circles
```
geom3trik, line 61: Unnecessary comment

examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+    circles: Vec<Circle>,
+    /// index of current selected circle
```
geom3trik, line 63: Unnecessary comment

examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+    Redo,
+    ToggleRightMenu,
+    ToggleDialog,
+}
+
+impl CircleDrawerData {
+    fn update_selected(&mut self, x: f32, y: f32) {
+        self.circles_data.selected =
+            self.circles_data.circles.iter().position(|c| distance(c.x, c.y, x, y) < c.r);
+    }
```
geom3trik, line 116: This will find the first circle where the mouse is within the radius, but what about overlapping circles? In this case it might find the wrong circle as you'd expect it to select the circle 'on top', which would be *after* the first one found.

geom3trik: One easy solution would be to add a `rev()` to the iterator so you're searching front to back.

examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+impl Model for CircleDrawerData {
+    fn event(&mut self, cx: &mut EventContext, event: &mut Event) {
+        if let Some(event) = event.take() {
+            match event {
+                CircleDrawerEvent::AddCircle(x, y) => {
+                    let circle = Circle { x, y, r: 26.0 };
+                    self.circles_data.selected = Some(self.circles_data.circles.len());
+                    self.circles_data.circles.push(circle);
+                    self.undo_list.push(UndoRedoAction::Circle(circle));
+                    self.redo_list.clear(); // we need to clear redo list in each new event
```
geom3trik, line 128: Why is this required?

examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+                    }
+                }
+                CircleDrawerEvent::ChangeRadius(r) => {
+                    if let Some(idx) = self.circles_data.selected {
+                        self.circles_data.circles[idx].r = r;
+                    }
+                }
+                CircleDrawerEvent::Undo => {
+                    let last = self.undo_list.remove(self.undo_list.len() - 1);
+
```
geom3trik, line 142: You could use `pop` here as you have for the circles vec later on.

examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+                            self.redo_list.push(UndoRedoAction::RadiusChange(
+                                idx,
+                                self.circles_data.circles[idx].r, // store the current radius in redo list
+                            ));
+                            self.circles_data.circles[idx].r = r // update the radius to the old one
+                        }
+                    }
+                }
+                CircleDrawerEvent::Redo => {
+                    let last = self.redo_list.remove(self.redo_list.len() - 1);
```
geom3trik, line 158: Same as above, could use `pop()`.

examples/7GUIs/circle_drawer.rs
```diff
@@ -0,0 +1,358 @@
...
+                                idx,
+                                self.circles_data.circles[idx].r, // store the current radius in undo list
+                            ));
+                            self.circles_data.circles[idx].r = r // update the radius
+                        }
+                    }
+                }
+                CircleDrawerEvent::ToggleRightMenu => {
+                    let x = cx.mouse().cursorx;
+                    let y = cx.mouse().cursory;
```
geom3trik, line 176: It probably doesn't matter here but it might be better to get the pressed position of the mouse button rather than the current position. You can do this with `cx.mouse().left.pos_down`.

geom3trik (approved):

examples/7GUIs/circle_drawer.rs
geom3trik: My main concern would be that the mouse might move before this event is processed, causing the menu to appear not where the click occurred but where the mouse currently is. It's probably unlikely to happen, but using the click position rather than current position would fix this. Is it left click or right click that opens the menu? You can also do `cx.mouse().right.pos_down` to get the mouse position when the right button was last pressed.

examples/7GUIs/circle_drawer.rs
geom3trik: Ah yeh makes sense. I was mainly referring to why the redo list is reset but that all makes sense.

geom3trik (approved):

PR #373, opened 2023-07-13
Add logging

geom3trik:


perlindgren (commented):
There is currently a security issue with fern. A quick fix is to add:

--features no-color 

to the dependency.

geom3trik:
> There is currently a security issue with fern. A quick fix is to add:
> 
> --features no-color
> 
> to the dependency.

I removed the `color` feature already. It now passes `cargo audit`

Issue #381, opened 2023-07-18
Miscellaneous issues and feature requests


geom3trik:
- [x] Ability to style textbox placeholder text
- [x] Show textbox placeholder text when the textbox is being edited but has no content
- [x] Ability to add preceding icon to textbox

PR #382, opened 2023-07-18
Textbox improvements

geom3trik:
- [x] Allow placeholder text to be styled
- [x] Show placeholder text when textbox is being edited but is empty
- [x] Add example for preceding icon
- [x] Use lens target as validation type to remove the need for the user to manually parse
- [x] Add `on_cancel` callback which is triggered when the escape key is pressed
- [x] Update textbox theming
- [x] Cleanup textbox code
- [x] Use command key for cut/copy/paste/select all on macOS


geom3trik:
Closes #381

PR #395, opened 2023-07-25
Change `Entity` and other generational IDs to use `u64` instead of `u32`

geom3trik:


geom3trik:
I suspect that there might be an underlying issue with the way generational IDs are created/destroyed or reused. However, until I have sufficient time to dig deep into this, I'm going to merge this for now as it's a valid fix for the problem. The problem being that in extreme circumstances, such as a virtual list with many items, we're running out of entity IDs.

PR #398, opened 2023-07-31
Create a custom_view example.

isaacmclellan:


geom3trik (changes requested):
This looks good so far but it should also include using the custom view in a `main` function so that the example can be run.

isaacmclellan:
> This looks good so far but it should also include using the custom view in a `main` function so that the example can be run.

It should be good now, I completely forgot about a main function lol.

geom3trik:
> > This looks good so far but it should also include using the custom view in a `main` function so that the example can be run.
> 
> It should be good now, I completely forgot about a main function lol.

You appear to have added the main function twice now.

geom3trik (approved):

PR #400, opened 2023-08-01
Minor improvements to the `Dropdown` view.

Greatness7:
1. Move styling to `default_layout.css` rather than hard-coding it in rust. This is needed to enable user customization.

2. Add `checked` modifier to the dropdown title. This allows it to be styled differently when opened versus closed.

It was suggested on discord that refactoring the `Dropdown` itself to hold the `is_open` state may be a better solution to `2`, but that is a larger change and I'm not familiar enough with the internals yet to implement it.

### DIFF
--- a/crates/vizia_core/resources/themes/default_layout.css
+++ b/crates/vizia_core/resources/themes/default_layout.css
@@ -171,6 +171,8 @@ dropdown .title {
 dropdown popup {
     width: 100%;
     height: auto;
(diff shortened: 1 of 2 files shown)


geom3trik (approved):
I'm happy with this for now. I think in future I might revisit the dropdown view and move the `is_open` flag to the view itself, but for now this is fine.

PR #412, opened 2023-08-09
add .editorconfig

isaacmclellan:


geom3trik:
I'm not familiar with `.editorconfig`, what's the purpose of this?

isaacmclellan:
> I'm not familiar with `.editorconfig`, what's the purpose of this?

https://editorconfig.org/

geom3trik:
If we could match this to what rust-analyser already does then I'd be happy to merge this

PR #419, opened 2023-10-04
Added `WindowEvent::ThemeChanged` event

YouKnow-sys:
## Added a new event named `ThemeChanged` to `WindowEvent`
This event will get emitted when current system theme changed.
I also added a new example to showcase this new event.
currently this new event is only supported with winit backend (Im not sure if baseview support such a thing).

### todo
- [x] add a new method to vizia to receive current system theme.

### DIFF
--- a/examples/helpers/mod.rs
+++ b/examples/helpers/mod.rs
@@ -1,14 +1,5 @@
 #![allow(dead_code)]
-use vizia::{
-    icons::{ICON_MOON, ICON_SUN},
-    prelude::*,
-};
-
-use lazy_static::lazy_static;
-
-lazy_static! {
-    pub static ref THEME_MODES: Vec<&'static str> = vec!["Light", "Dark"];
-}
+use vizia::prelude::*;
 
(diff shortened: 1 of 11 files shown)


geom3trik (commented):
This is looking good to me so far.

examples/helpers/mod.rs
```diff
@@ -165,3 +135,12 @@ impl ExamplePage {
 }
 
 impl View for ExamplePage {}
+
+fn theme_selection_dropdown(cx: &mut Context) {
+    PickList::new(cx, ControlsData::theme_options, ControlsData::selected_theme, true)
+        .on_select(|cx, index| cx.emit(ControlsEvent::SetThemeMode(index)))
```
geom3trik, line 141: Just a thought. Instead of emitting a `ControlsEvent` here and then handling that and translating it to an `EnvironmentEvent::SetThemeMode` event. Could do that here instead and cut out the middle man?

geom3trik: Ah yes good point! Missed that part

Issue #422, opened 2023-10-12
Mouse Position Events Issue


ejjonny:
Branch: [baseview-update](https://github.com/vizia/vizia/tree/baseview-update)
Repro: 
Add a label to the knob example
```rust
fn main() {
    Application::new(|cx| {
        AppData { value: 0.2 }.build(cx);

        ExamplePage::new(cx, |cx| {
            VStack::new(cx, |cx| {
                Knob::new(cx, 0.5, AppData::value, false).on_changing(|cx, val| {
                    cx.emit(AppEvent::SetValue(val));
                });
                Label::new(cx, AppData::value);
            });
        });
    })
    .title("Knob")
    .inner_size((250, 250))
    .run();
}
```
Run the example using baseview
`cargo run --example knob --no-default-features --features baseview`

Observe stuttering / jumping mouse position updates while dragging the knob.

https://github.com/vizia/vizia/assets/17223924/b32b0c4e-9e17-44ca-b749-df935b8cd56a

This event published in vizia_core/src/systems/layout.rs line 140 is part of the problem in some way, the issue doesn't occur if this code is removed. 

```rust
        // A relayout, retransform, or reclip, can cause the element under the cursor to change. So we push a mouse move event here to force
        // a new event cycle and the hover system to trigger.
        if let Some(proxy) = &cx.event_proxy {
            let event = Event::new(WindowEvent::MouseMove(cx.mouse.cursorx, cx.mouse.cursory))
                .target(Entity::root())
                .origin(Entity::root())
                .propagate(Propagation::Up);

            proxy.send(event).expect("Failed to send event");
        }
```

Using vizia with nih-plug this event seems to overwrite the real window mouse event if layout is being called every frame. If both this event & the window mouse event publish once per frame & this event is processed after the real mouse event, it perpetuates an old mouse position event. I think the issue seen in the knob example here is likely similar.

PR #423, opened 2023-10-12
Baseview mouse event fix

ejjonny:
Possible fix for #422 

Prevents this sequence of events:

1. Frame update calls layout system
2. Baseview sends mousemove & layout system sends a mousemove
3. Frame update processes baseview mousemove, & then processes layout mousemove overwriting the real cursor position
4. Frame update then calls layout which re-sends the stale cursor value for the next frame update
5. Repeat from 3 for the duration of the gesture triggering relayout, or until relayout is not performed on a frame

An alternative fix may be to prevent such frequent layout, or to investigate what is different in winit feature that prevents this, but I think the mousemoved event that was being sent here was sort of a band-aid in the first place?

### DIFF
--- a/crates/vizia_baseview/Cargo.toml
+++ b/crates/vizia_baseview/Cargo.toml
@@ -12,7 +12,7 @@ vizia_core = { path = "../vizia_core" }
 vizia_input = { path = "../vizia_input" }
 vizia_id = { path = "../vizia_id" }
 
-baseview = { git = "https://github.com/RustAudio/baseview.git", rev = "c129b12ead4f5ac02126f559ceb8ce43cc982200", features = ["opengl"] }
-raw-window-handle = "0.4.2"
+baseview = { git = "https://github.com/RustAudio/baseview.git", rev = "f0639b787bbda506434d3f6b5c91e94ca59904c6", features = ["opengl"] }
+raw-window-handle = "0.5.2"
 femtovg = "0.7.0"
 lazy_static = "1.4.0"
--- a/crates/vizia_baseview/src/application.rs
+++ b/crates/vizia_baseview/src/application.rs
@@ -213,7 +213,6 @@ impl ApplicationRunner {
         while let Some(event) = queue_get() {
             cx.send_event(event);
         }
-
         // Events
         cx.process_events();
 
--- a/crates/vizia_baseview/src/application.rs
+++ b/crates/vizia_baseview/src/application.rs
@@ -247,7 +247,7 @@ impl ApplicationRunner {
         }
 
         // Force restyle on every frame for baseview backend to avoid style inheritance issues
-        cx.style().needs_restyle();
+        // cx.style().needs_restyle();
         cx.process_data_updates();
 
         let context = window.gl_context().expect("Window was created without OpenGL support");
@@ -343,6 +343,15 @@ impl ApplicationRunner {
 
                     cx.emit_origin(WindowEvent::MouseScroll(lines_x, lines_y));
                 }
+
+                baseview::MouseEvent::CursorEntered => {
+                    cx.emit_origin(WindowEvent::MouseEnter);
+                }
+
+                baseview::MouseEvent::CursorLeft => {
+                    cx.emit_origin(WindowEvent::MouseLeave);
+                }
+
                 _ => {}
             },
             baseview::Event::Keyboard(event) => {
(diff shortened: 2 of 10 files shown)


geom3trik:
It's not a band-aid as such because it happens for a specific reason. A mouse move event triggers the hover system to determine which view is under the mouse. This has to be done after layout (if required). So its mouse move -> layout -> hover. However, if hovering a view causes it to move, the layout system needs to be called again and the hover system should be called again to determine what view is now under the mouse. The easiest way to do this is to push another mouse move event into the stack.

I'm not sure what the exact issue is right now, but something we did for the winit backend and possibly not the baseview one is to debounce the raw mouse move events. So if we get multiple mouse move events before an events cleared we just take the last one.

ejjonny:
Oh yeah, I see now, winit has some state in the event loop to track whether we've moved the mouse already in a given iteration of the event loop.
Maybe it would still be useful to define an event specifically for this purpose rather than overloading mouse move events, but I don't know the repercussions well enough to be sure! 
I imagine the distinction between 'the mouse moved' & 'my view moved under the mouse' could be useful to someone implementing a view, or it could just be a pain to respond to both.

geom3trik:
Yeh we could have another event for it I guess but the end result is the same. I.e. run the layout and hover systems again within the same loop.

I think the problem here is that baseview doesn't have a `MainEventsCleared` equivalent. The mouse move debounce was put in because a high rate mouse would cause multiple mouse move events per one main events cleared. I suspect the baseview backend will just fire the events function for every mouse move. 

I'll take a proper look at your PR soon but I think you're right that it's conflicting mouse positions from different sources within the same frame.

geom3trik:
I'm not sure about introducing a new event just for this, but I assume this works because it's not overriding the mouse position in context and instead using what's already been set by the mouse move event?

ejjonny:
Yeah. I looked a bit more today, the reason the knob moves wildly is because it renders based on mouse movement delta across frames. Since the stale mouse position continues to publish between real mouse move events the knob component processes a delta that doesn't exist & moves backwards or jumps around.

With some logging you can see the 307.3... position strung along over time.

```
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 307.34766
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 309.16016
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 307.34766
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 309.8203
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 307.34766
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 310.22656
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 310.6211
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 307.34766
[crates/vizia_core/src/events/event_manager.rs:191] context.mouse.cursory = 311.01953
```

geom3trik:
Is there a way to fix this without introducing a new window event?

ejjonny:
The best alternative I can think of would be to add some 'needs_hover' state somewhere appropriate.

Hover ultimately depends on layout, but layout happens after events are processed, so you could set `needs_hover = true` while processing events & then call the hover system once layout is ready.
This should make the mousemove event published during layout unnecessary.

This way nothing is delayed to the next frame, no events are overloaded, and no new events are introduced.

This is assuming layout doesn't directly or indirectly depend on hover being finished, I'm not sure whether that's the case.

If this sounds reasonable, I'd like your thoughts on the best place for the needs_hover state.

geom3trik:
> The best alternative I can think of would be to add some 'needs_hover' state somewhere appropriate.
> 
> Hover ultimately depends on layout, but layout happens after events are processed, so you could set `needs_hover = true` while processing events & then call the hover system once layout is ready.
> This should make the mousemove event published during layout unnecessary.
> 
> This way nothing is delayed to the next frame, no events are overloaded, and no new events are introduced.
> 
> This is assuming layout doesn't directly or indirectly depend on hover being finished, I'm not sure whether that's the case.
> 
> If this sounds reasonable, I'd like your thoughts on the best place for the needs_hover state.

The issue isn't really that the hover system needs to happen after layout, the extra event was necessary to wake up the event loop and trigger another run of the hover system. For example:

1. User moves the mouse over a view
2. Hover system is run (I'm assuming layout has already happened previously).
3. Style system runs and determines that the view should be hidden/moved when hovered.
4. Layout runs and moves the view. The mouse is now hovering over the view underneath the one that moved.
5. The event loop waits for an event but a moving view doesn't trigger one (well it technically emits a geometry changed event)
6. Push a manual mouse move event to wake the event loop and trigger the hover system to run.
7. Hover system determines the view underneath is the one actually being hovered.

Issue #247, opened 2022-10-06
calling .close() on window handle doesn't free Application (vizia-baseview)


sebastianpfluegelmeier:
when calling `window_handle::close()` on the `WindowHandle` you get from `application.open_parented()` the application in question doesn't get freed. This applys to vizia-baseview. I did not test it with other implementations.
This leaks memory, especially with audio plugins where each time you open your plugin UI, you create a new Application.

sebastianpfluegelmeier:
so i did some experiments to confirm that vizia is actually leaking

```rust
use vizia_baseview::Application;

fn main() {
    loop {
        Application::new(move |_cx| {})
            .open_as_if_parented()
            .close();
    }
}
```

this is continously leaking memory, even though each Loop the application should get freed

sebastianpfluegelmeier:
i think the problem comes from baseview, i tried this now
```rust
use baseview::{Event, EventStatus, Window, WindowHandler, WindowScalePolicy};

struct MemoryLeakExample {}

impl WindowHandler for MemoryLeakExample {
    fn on_frame(&mut self, _window: &mut Window) {}

    fn on_event(&mut self, _window: &mut Window, _event: Event) -> EventStatus {
        EventStatus::Ignored
    }
}

fn main() {
    loop {
        let window_open_options = baseview::WindowOpenOptions {
            title: "memory leak example".into(),
            size: baseview::Size::new(100.0, 100.0),
            scale: WindowScalePolicy::SystemScaleFactor,
        };
        Window::open_as_if_parented(window_open_options, |_| MemoryLeakExample {}).close();
    }
}
```
and it also leaks memory continously. should i close the issue or leave it open as it still persists even though it has to be fixed in baseview?

geom3trik:
I would say leave the issue open to give visibility to the bug which can affect vizia users. If you end up making an issue on the baseview repo then you can link it here.

sebastianpfluegelmeier:
as discussed in the discord, `open_as_if_parented()` leaks memory by design, but if you use `open_parented()` there is a memory leak as well and this seems to be caused by openGL.

PR #441, opened 2023-12-18
Update Baseview

glowcoil:
Update to the latest version of Baseview. Includes some important fixes for window lifetime management on macOS and removes the `Window::open_as_if_parented` API.


geom3trik:
Does nih_plug use `open_parented` rather than `open_as_if_parented` then do you know?

glowcoil:
Yeah, NIH-plug doesn't actually make use of `open_as_if_parented` anywhere, and as far as I know nothing else does either. It's specifically only useful for Audio Unit plugins on macOS (using it in other contexts is incorrect).

geom3trik:
Fixes #247

geom3trik (approved):

PR #443, opened 2023-12-30
Avoid accidentally changing the logical size when changing user scale due to rounding errors

robbert-vdh:
This happened due to rounding errors in the conversion to and from integers. We can detect this by doing the same computation, and treating the resize as a noop if it does happen. This prevents the logical size from slowly drifting when changing the user scale many times in a row.

### DIFF
--- a/crates/vizia_baseview/src/application.rs
+++ b/crates/vizia_baseview/src/application.rs
@@ -168,7 +168,8 @@ pub(crate) struct ApplicationRunner {
     /// events then the window will be resized.
     current_user_scale_factor: f64,
     /// The window's current logical size, before `user_scale_factor` has been applied. Needed to
-    /// resize the window when changing the scale factor.
+    /// resize the window when changing the scale factor. Can also be changed through external
+    /// events.
     current_window_size: WindowSize,
 }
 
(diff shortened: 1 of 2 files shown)


crates/vizia_baseview/src/application.rs
```diff
@@ -207,26 +208,17 @@ impl ApplicationRunner {
...
                 width: self.current_window_size.width as f64 * self.current_user_scale_factor,
                 height: self.current_window_size.height as f64 * self.current_user_scale_factor,
             });
 
-            // TODO: These calculations are now repeated in three places, should probably be moved
-            //       to a function
+            // TODO: Without this `WindowEvent::GeoChanged` isn't emitted for every element, even
+            //       though this same function is also called in the
+            //       `baseview::WindowEvent::Resized` event handler. Why?
             cx.set_scale_factor(self.window_scale_factor * self.current_user_scale_factor);
```
robbert-vdh, line 221: This is the part I don't get. This call shouldn't be necessary since the event handler calls the same function with the same values, but without it not all `GeometryChanged` events are fired. Specifically, the event is not fired for `Model`s, like in the `user_scale` example.

robbert-vdh: Something similar happens when changing just the window size (and this call doesn't help there). `GeometryChanged` isn't called for `Model`s when changing just the window size.

robbert-vdh:
I'll probably end up changing the `Window::resize` semantics in baseview (so it's consistent between platforms, which it isn't right now) and then I'll update Vizia to match.

PR #445, opened 2024-01-15
Added the new `Custom` variant to AppTheme

YouKnow-sys:
using this new variant user can set their own custom theme for vizia to use as internal theme.
some other small changes:
- `Theme::get_current_theme` will now return a option instead of the `ThemeMode` directly,
- `EventContext::set_theme_mode` and the newly added `EventContext::set_custom_theme` are now `pub(crate)` because before users were able to change the theme directly using these methods, by doing this the `AppTheme` in env could get out of sync compared to the current theme. now we recommend the users to change the theme using `EnvironmentEvent::SetThemeMode` event.

We could also add the same hot reload functionality that Vizia support for user themes to the new `Custom` variant if needed.

### DIFF
--- a/crates/vizia_core/src/context/mod.rs
+++ b/crates/vizia_core/src/context/mod.rs
@@ -570,7 +570,7 @@ impl Context {
                         ThemeMode::DarkMode => self.add_theme(DARK_THEME),
                     }
                 }
-                AppTheme::Custom(ref theme) => self.add_theme(theme),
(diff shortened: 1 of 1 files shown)


geom3trik:
So does this work by overriding the built-in theme?

geom3trik:
Yeh I think having some way to 'add' to the existing themes, but still have them be selectable with the environment event would be good.

PR #451, opened 2024-01-28
RFC: Make VirtualList rebuild logic more "fine-grained".

Greatness7:
I had the observation that `VirtualList` currently recreates the entire set of visible items when the scroll position changes. In practice, at least for my app, it's usually only one or two items that pass over the ends of the list on any particular frame of scrolling. So in theory I only need to rebuild those one or two items and not the whole thing (60+ rows in my case).

This PR is my attempt at implementing that idea. 

The general idea of the implementation goes something like this:

* The virtual list holds a `visible_items` list that contains the indices of the items in the backing list that are currently visible.

* For example, `[0,1,2,3]`, assuming you're at the initial scrollbar position with enough screen space for 4 items.

* We create an associated container for each slot, with bindings inside to monitor the index and rebuild the user content when changes occur.
```rs
HStack::new(cx, |cx| {
    Binding::new(cx, visible_items.index(N), |cx, lens| {
        let index = lens.get(cx);
        let item = list.index(index);
        user_content(cx, index, item);
    });
})
```
* Now, the only magic that needs to be done is to make sure that as we scroll through the list we only update as few of the indices as possible, thus ensuring that our bindings don't trigger unnecessarily.

* So, if you scroll forward one item, you now have `[1,2,3,4]` in view. The `0` dropped off the front and `4` was introduced at the back. We want to preserve the `[1,2,3]` section, so we just put the `4` where the `0` used to be and get `[4,1,2,3]`. Now only one of our bindings detects a change and triggers a rebuild, rather than the whole entire list as would happen before.

The amount of performance gained is going to depend on how costly the virtual list rows are to rebuild. The `virtual_list` example in the repo only uses a single label so it does not benefit much. In my own app, when at full screen, around 60+ rows are visible, and each one is a composition various labels, icons, etc. My FPS while scrolling through them more than doubles when using this PR (from around ~60 to ~140). I have a relatively low resolution monitor, so 4k users might have lists of 100+ items on screen and benefit even more.

I suppose a downside to this approach is that you have all these bindings running in the background, though they're only looking at a single `usize` each so I doubt it matters. In theory the indices only change when appropriate events happens so maybe there's a better way to do that? It may also be possible to not use a list at all for tracking visible items and just use some kind of (start, stop, partition_point)...

Another downside is that the visible items no longer cleanly represent a `Range` in the internal data structure, so the `on_change` callback has to do a little more work, maybe it shouldn't return a `Range` at all?

Finally, I had to add a `Copy` bound on the user content argument. I'm not sure the implications of this? I believe `Handle` are already `Copy` so maybe it's meaningless? The bound does not seem to affect any of my usages of `VirtualList`, and my app is essentially just a whole bunch of `VirtualList` all over the screen.

### DIFF
--- a/crates/vizia_core/src/views/virtual_list.rs
+++ b/crates/vizia_core/src/views/virtual_list.rs
@@ -1,11 +1,11 @@
-use std::ops::Range;
+use std::ops::{Deref, Range};
 
 use crate::prelude::*;
 
 #[derive(Lens)]
 pub struct VirtualList {
-    offset: usize,
     num_items: usize,
+    item_offset: usize,
     item_height: f32,
     visible_items: Vec<usize>,
     scrolly: f32,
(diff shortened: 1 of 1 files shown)


crates/vizia_core/src/views/virtual_list.rs
```diff
@@ -19,75 +19,120 @@ pub enum VirtualListEvent {
...
 
 impl VirtualList {
-    pub fn new<V: View, L, T>(
+    pub fn new<V: View, L: Lens, T: 'static>(
         cx: &mut Context,
         list: L,
-        height: f32,
-        item: impl Fn(&mut Context, usize, Index<L, T>) -> Handle<V> + 'static,
+        item_height: f32,
+        item_content: impl Fn(&mut Context, usize, Index<L, T>) -> Handle<V> + Copy + 'static,
```
geom3trik, line 26: The `Copy` constraint is due to having to move the content closure into the binding closure. I *think* this is fine, but if at some point we find this isn't fine then a way around it is to wrap it in `Arc` and clone the arc into the closure.

geom3trik (approved):
This looks good to me. Tested it out and it works well.

PR #459, opened 2024-02-09
Add deps.rs badge to README

uklotzde:
I have noticed that many dependencies are outdated. Upgrading them requires migrating existing code. Relying on unmaintained versions is not an option.

This PR adds a [deps.rs](https://deps.rs/repo/github/vizia/vizia) badge to the README to show the current status.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,11 @@
       alt="CI status" />
   </a>
   <!-- docs (TODO) -->
+  <!-- Dependency status -->
+  <a href="https://deps.rs/repo/github/vizia/vizia">
+    <img src="https://deps.rs/repo/github/vizia/vizia/status.svg"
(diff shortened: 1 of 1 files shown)


geom3trik:
Happy to add this badge, but just note that some of the code changes required to update these dependencies is quite substantial. For example, the winit update brought a lot of API breaking changes that we're still trying to work through.

uklotzde:
> ...some of the code changes required to update these dependencies is quite substantial.

Sure. I noticed this too. All this does not need to be fixed immediately. But we should be aware of it.

Keeping pace with the building blocks is essential, especially during the early phase. Those breaking changes might even affect architectural decisions. Ignoring them for too long could be fatal.

geom3trik (approved):

PR #463, opened 2024-03-19
Set input focus on alt-click

dathinaios:
I have submitted a PR for the input focus issue (https://github.com/RustAudio/baseview/pull/170). Assuming it is accepted and merged in `baseview` these are the changes required on `vizia_baseview`.

I have opted for the least intrusive call on alt-click but I wonder if there are other views that would need it. Let me know what you think.

I have tested this as mentioned in the PR comments.

### DIFF
--- a/crates/vizia_baseview/src/application.rs
+++ b/crates/vizia_baseview/src/application.rs
@@ -292,6 +292,13 @@ impl ApplicationRunner {
                 baseview::MouseEvent::ButtonPressed { button, modifiers } => {
                     update_modifiers(modifiers);
 
+                    // give input focus to the view on alt-click
+                    if modifiers == vizia_input::KeyboardModifiers::ALT {
+                        if !window.has_input_focus() {
+                            window.set_input_focus();
+                        };
+                    };
+
                     let b = translate_mouse_button(button);
                     cx.emit_origin(WindowEvent::MouseDown(b));
                 }
--- a/crates/vizia_baseview/src/window.rs
+++ b/crates/vizia_baseview/src/window.rs
@@ -175,10 +175,10 @@ impl WindowHandler for ViziaWindow {
         unsafe { context.make_not_current() };
     }
 
-    fn on_event(&mut self, _window: &mut Window<'_>, event: Event) -> EventStatus {
+    fn on_event(&mut self, window: &mut Window<'_>, event: Event) -> EventStatus {
         let mut should_quit = false;
-        self.application.handle_event(event, &mut should_quit);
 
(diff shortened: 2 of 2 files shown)


geom3trik:
I've opened #466 which allows window events to be handled by baseview. I think this could be used in conjunction with what you have here to focus the window when requested by say a textbox being focused etc.

dathinaios:
> I've opened #466 which allows window events to be handled by baseview. I think this could be used in conjunction with what you have here to focus the window when requested by say a textbox being focused etc.

Are you thinking of adding something like this:

```rust
 ViziaWindow { application, on_idle, should_close: false, should_focus: false }
 ```
 
 and then handling the event similarly to what you did for close?

geom3trik:
> > I've opened #466 which allows window events to be handled by baseview. I think this could be used in conjunction with what you have here to focus the window when requested by say a textbox being focused etc.
> 
> Are you thinking of adding something like this:
> 
> ```rust
>  ViziaWindow { application, on_idle, should_close: false, should_focus: false }
> ```
> 
> and then handling the event similarly to what you did for close?

No, that's only there for close because closing the window needs to happen after event handling. What I'm proposing is adding a focus event to `WindowEvent`, which can be emitted by a textbox and then handled by the callback I've added to the process events function so that the relevant function can be called on the baseview window.

dathinaios:
> No, that's only there for close because closing the window needs to happen after event handling. What I'm proposing is adding a focus event to `WindowEvent`, which can be emitted by a textbox and then handled by the callback I've added to the process events function so that the relevant function can be called on the baseview window.

Oh, I see. So I will just have to add the switch inside `handle_event` in its own event handler right? That should be find though we still need to pass in the window.

geom3trik:
> Oh, I see. So I will just have to add the focus code inside `handle_event` in its own event handler right? That should be fine though we still need to pass in the window.

The focus code would go inside the closure I've added to `process_events` in the linked PR. You can pass the window into this and call focus on it.

dathinaios:
> The focus code would go inside the closure I've added to `process_events` in the linked PR. You can pass the window into this and call focus on it.

Ok, I will have a look thank you. Unfortunately at the moment `vizia` and `nih_plug` are out of sync. I still need to use [Robbert's Vizia fork](https://github.com/robbert-vdh/nih-plug/blob/bb274976ca138289436b0bfa209b2c8c13244225/nih_plug_vizia/Cargo.toml#L17) locally to test the input focus fix. Once the changes are merged in `nih_plug_vizia` I can implement and test.

PR #464, opened 2024-03-20
Temporary fix for the default font being too bold in skia.

Greatness7:
Fixes bold text and icons being too thick (especially at small sizes). This is a temporary solution until font variation parameters are configurable.


geom3trik:
Does this still work if the font is not variable?

Greatness7:
> Does this still work if the font is not variable?

It does not. Hence the 'temporary solution' bit. Ideally these will be configurable in CSS before `skia` makes it into `main`. This was just to have less-irritating text rendering in the meantime (for the default font). Without this commit `semi-bold` and other such options are not working currently, and the default bold that appears in their place is very extreme.

PR #465, opened 2024-03-23
Add support for configuring 'font-variation-settings' in CSS.

Greatness7:
```rs
use vizia::prelude::*;

const STYLE: &str = r#"
    :root {
        font-size: 30;
        font-variation-settings: 'wght' 700.0, 'slnt' -10.0, 'wdth' 150.0;
    }
"#;

fn main() {
    Application::new(|cx| {
        cx.add_stylesheet(STYLE).unwrap();
        Label::new(cx, "Hello World!");
    })
    .run()
    .unwrap();
}
```

Reference: <https://developer.mozilla.org/en-US/docs/Web/CSS/font-variation-settings>

I've only implemented the explicit syntax. AKA the `'wght' 400` form. The mozilla page mentions other forms with keywords like `revert-layer`, but I wasn't sure if those are applicable for vizia.

This is the first time I've looked at CSS parts of the code base. I tried to emulate what the other font settings were doing, and it seems to work, but it's worth reviewing in case I've done something wrong.


crates/vizia_core/src/systems/text.rs
```diff
@@ -73,79 +72,91 @@ pub fn build_paragraph(
...
+        ) {
+            (None, None, None) => {}
+            (weight, width, slant) => {
+                text_style.set_font_style(FontStyle::new(
+                    weight.copied().unwrap_or_default().into(),
+                    width.copied().unwrap_or_default().into(),
+                    slant.copied().unwrap_or_default().into(),
+                ));
+            }
+        }
```
geom3trik, line 145: Not sure I understand this part. Wouldn't this only set the font style if all of weight, width, and slant are defined?

Greatness7: No, the idea was that it will not set a font style if *all* of them are `None`. If any are not `None` then it will take the 2nd match.

Partially as an optimization, but mainly I thought it would be best to avoid applying the font style unless its explicitly requested. To try and not conflict with variation parameters. 

Though now that I look at the MSDN page, it seems that the correct thing to do is have it so variations always [override the equivalent font style](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variation-settings#description). Though I'm not sure what the unit conversions should be in that case (like at what threshold of `slnt` do you become oblique/italic?).

geom3trik: Ah yes of course. Do we know whether font arguments override the equivalent font style in skia? I thought they did but I can't remember

Greatness7: I just tested and skia does not handle this for us unfortunately.
```cs
    font-weight: bold;
    font-variation-settings: 'wght' 320;
```
These combined together look different than either one does in isolation.

geom3trik (approved):

PR #470, opened 2024-04-07
Feat: Adds xypad and an example for it

bboettcher3:
* Adds XY pad view
    * Supports disabling
    * Supports light/dark themes
* Adds an example for the new view
Screenshot 2024-04-07 at 1 41 44 PM
Screenshot 2024-04-07 at 11 41 26 AM

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -256,6 +256,10 @@ name = "tabview"
 path = "examples/views/tabview.rs"
 
 [[example]]
+name = "xypad"
+path = "examples/views/xypad.rs"
+
+[[example]]
 name = "circle_drawer"
 path = "examples/7GUIs/circle_drawer.rs"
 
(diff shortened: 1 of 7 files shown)


bboettcher3:
One odd behavior is that the xypad element and thumb CSS classes don't change the border properties I want it to on dark theme.. Maybe someone could check if I'm missing something there? Other properties like background-color change correctly, but the border-color has no effect.

geom3trik:
> One odd behavior is that the xypad element and thumb CSS classes don't change the border properties I want it to on dark theme.. Maybe someone could check if I'm missing something there? Other properties like background-color change correctly, but the border-color has no effect.

It looks like you've set a border colour for it in rust code. This will override any css styling on the element.

bboettcher3:
> It looks like you've set a border colour for it in rust code. This will override any css styling on the element.

Ahh, fixed that and updated the image above with the new theme, thanks!

geom3trik (approved):

magnetophon:
I totally missed this merge!
Thanks, @bboettcher3 and @geom3trik!

PR #472, opened 2024-04-12
Fixes for circle demo

GammaSagittarii:
Reverted to using popup widget instead of dialog button, fixed menu not sending on_press event, and displaying coordinates in the popup for the circle radius size.

### DIFF
--- a/examples/7GUIs/circle_drawer.rs
+++ b/examples/7GUIs/circle_drawer.rs
@@ -1,5 +1,6 @@
 use vizia::prelude::*;
-use vizia::vg::{Paint, Path};
+use vizia::vg::{Paint, PaintStyle, Path, Point};
+use vizia_winit::application::{Application, ApplicationError};
 
 const STYLE: &str = r#"
     :root {
@@ -193,8 +194,8 @@ impl Model for CircleDrawerData {
                 }
 
                 if !self.dialog_open {
-                    let x = cx.physical_to_logical(cx.mouse().cursorx);
-                    let y = cx.physical_to_logical(cx.mouse().cursory);
+                    let x = cx.physical_to_logical(cx.mouse().cursor_x);
+                    let y = cx.physical_to_logical(cx.mouse().cursor_y);
 
                     self.circles_data.update_selected(x, y);
                 }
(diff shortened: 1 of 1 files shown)


geom3trik:
Looking good! However there is a scaling issue. The mouse coordinates are physical pixels but the layout system works in logical. You just need a `cx.physical_to_logical()` before the `Pixels()` parts.

GammaSagittarii:
Thank you, that is fixed now. But that made me realize that the circle position is reported in scaled coordinates.
Is there an easy way to get EventContext from Context?
I was unable to figure out how to call physical_to_logical in the map closure, as cx here is Contex and not EventContext in:
`CircleDrawerData::circles_data
                                        .then(CircleData::circles)
                                        .get(cx)
                                        .get(selected)
                                        .map(|c| (c.x, c.y))
                                        .unwrap()`

GammaSagittarii:
I've figured out a workaround, by adding fields, x_logical and y_logical  to the Circle, I can keep track of physical_to_logical values, and display that data. I think that adds unnecessary complexity to the demo, and displaying scaled coordinates is probably of no value anyway. Let me know what you think.

GammaSagittarii:
No need to apologize, your timing is perfect. Before trying my workaround I tried to store coordinates in logical space but failed and got a nice app panic. I guess I was just having a bad day, because I managed to do it now, without much trouble.

GammaSagittarii:
Hi, @geom3trik I've fixed this pull request to work with the latest master branch.
There is only one issue, and that is that undo and redo buttons disappear after opening the dialog for the circle resizing. They show up again if you put the cursor where they are supposed to be.

geom3trik:
> There is only one issue, and that is that undo and redo buttons disappear after opening the dialog for the circle resizing. They show up again if you put the cursor where they are supposed to be.

Thanks for updating this and pointing that out. Took a while to find but it turned out to be an overflow bug when computing clip intersections. I've fixed that now and the example seems to work as expected.

I'm just going to do some clean up to the old style rules that don't apply anymore, and then I think this is ready to merge. Thanks again :)

GammaSagittarii:
Yes, I suspected they should be removed, I had to use max-width properties to make the layout work.
Thank you!

geom3trik:
> I had to use max-width properties to make the layout work. 

That's because the popup has a min width of 100%, because it was originally designed to work with dropdowns. I've adjusted it in the CSS for now but I think in the future it might be worth having a separate view for popup menus. I'll have a think on this.

geom3trik (approved):

PR #492, opened 2024-08-25
Toggle `window.always_on_top` while app is running

cangzhang:
Handle `WindowEvent::SetAlwaysOnTop` and toggle window level while app is running


cangzhang:
Fixed clippy error

cangzhang:
@geom3trik updated, plz have a look

PR #501, opened 2024-09-04
Updated `Cargo.toml`

YouKnow-sys:
Previously we were hardcoding `vizia` path in examples `widget_gallery` and `designer` and that caused cargo to fail compiling the project if we had renamed the vizia directory to something other then `vizia`... this PR aim to fix this problem.
we now have the `vizia` crate available as a workspace dependency, so we can use it directly in mentioned examples.

# Changes
- put some common package information in workspace and used them in all crates when possible.
- put vizia itself inside workspace dependencies so we can use it internally, without hardcoding vizia path.
- commented out 'reqwest' as its unused right now.

# Questions
- [x] I noticed most of `vizia_*` crates have their own `rust-version` set, should we deleted them? imo its unnecessary.
- [x] should we update `rust-version` to `1.80`? were already using `1.79` and updating it to `1.80` could give us some useful things like the new stabilized `LazyCell` and `LazyLock`, with these we can potentially remove the `lazy_static` crate from vizia.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ version.workspace = true
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-rust-version = "1.79"
+rust-version = "1.80"
 
 [features]
 default = ["winit", "clipboard", "x11", "wayland"]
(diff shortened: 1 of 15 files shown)


geom3trik:
I'm not very good at this, so I'm happy to defer to your wisdom on the cargo stuff. Also happy to remove unnecessary Rust version stuff and bump the version to 1.8.

geom3trik (approved):

PR #503, opened 2024-09-05
Fix incorrect type used in `WindowEvent::SetSize` and related events.

Greatness7:
Fixes window size not being round-tripped correctly when being saved/loaded via `cx.window_size()` then `Application::inner_size()`. I believe it was previously working only on systems without DPI scaling; my windows desktop behaved as expected but macos laptop did not.

It seems to be the cause was that the `window_size()` method was retrieving the size from `cx.cache.get_bounds(parent_window)` which uses physical units rather than logical.

### DIFF
--- a/crates/vizia_winit/src/window.rs
+++ b/crates/vizia_winit/src/window.rs
@@ -381,17 +381,20 @@ impl View for Window {
             }
 
             WindowEvent::SetSize(size) => {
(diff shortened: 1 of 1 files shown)


geom3trik (approved):

arthurgeek:
with this change, using `inner_size((x, y))` in `Application` now uses `PhysicalSize`, which is incorrect for systems with DPI scaling.

PR #507, opened 2024-09-12
Fix for map_ref on already mapped data

geom3trik:


hydra:
this commit has been working well for me, no issues encountered.

PR #529, opened 2025-01-03
Add letter-spacing, word-spacing, line-height, and text-shadow properties + new text properties example

geom3trik:


geom3trik:
> There seems to be something off with regards to the bounding boxes of text with a negative letter spacing:

So it turns out that the reason the text was being cut off was because it was wrapping, but because wrapping was turned off it was clipping. I can prevent the wrapping so that words aren't clipped, but it seems that skia is incorrectly calculating text bounds when letter spacing is applied, and not just for negative letter spacing. If I remove the padding so that the bounds surround just the text,   some values of negative letter spacing cause the bounds to be too small, while some values of positive letter spacing cause the bounds to be too large (see images below). To be honest I'm not quite sure what to do about this as it might be a bug in skia.

image

image

PR #532, opened 2025-01-06
Make markdown (via comrak) an optional feature to save on deps

midnightveil:
Previously, `cargo build --release`: 985M target/ ,~1m50s build, ~300 deps

Now (default): ~1m30s build, 650M target, ~270 deps.

Could make this enabled by default, but at the very least it's nice to be able to turn it off.


geom3trik:
Comrak recently got a new release which removed some dependencies and made others optional: https://github.com/kivikakk/comrak/releases/tag/v0.33.0

I'm happy to make it optional still but I wonder what build times we get by just upgrading.

geom3trik (approved):

PR #537, opened 2025-01-12
Added 'legacy_x86_64_support' features to qfilter dependency.

Badless:
Added support for CPUs without popcnt instructions.


geom3trik:
As discussed on Discord, this can be fixed on the user's end by including the qfilter dependency with the required feature since features are additive. To avoid the performance hit to support older CPUs I'm going to recommend this solution.

Issue #534, opened 2025-01-08
Moveing between monitors having different scaling causes window shift on resize


ponchofiesta:
Having two monitors using different scaling causes issues when moving a window between them. A window resizes its physical size when moving between these monitors. In specific scenarios the window moves itself back to the previous monitor when the window resizes itself.

Here is an example to make this more clear:

- Left monitor has scaling 1.5
- Right monitor has scaling 1.75
- Move the window from right to left

Result:
As soon as the window has passed half to the left monitor, the window resizes itself and gets bigger. x and y position stays the same. The center of the window is now back to right monitor. When moving further to the left, the windows enlarges everytime the center of the window passes the monitors.

Possible solution:
- When moving from higher scaling factor to lower scaling factor the window physical size needs to shrink (actually it gets bigger).
- When resizing the window position needs to adapt to make the window center position stay on the same size. This is what other Windows programs behave (e.g. Windows Explorer and Calculator).

This may or may not be related to #522 But I think this is another problem.

https://github.com/user-attachments/assets/95f611c6-0d28-489a-9b23-adb9fe1464a4

geom3trik:
Sorry for the delay but this should be fixed now.

ponchofiesta:
Thank you. I just tested the current main revision on the timers demo. For me it didn't fix this issue. But now the window content is left aligned.

But the scaling of the content is looking good now. Maybe this fix was intended for #522 ?

geom3trik:
The left alignment a separate issue due to a new layout system. Both issues should be fixed on main now.

ponchofiesta:
Alignment is fixed again but the root issue here does still exist.

geom3trik:
Relevant winit PR: https://github.com/rust-windowing/winit/pull/4119

Issue #522, opened 2024-12-16
HiDPI scaling not working


ponchofiesta:
HiDPI scaling is not working. Actually the source code doc tells, this is handled by the `Resized` event. The relevant code is located in `baseview::WindowEvent::Resized` handler. But this event and even the whole `ApplicationRunner::handle_event` function is never called. Instead the `winit::event::WindowEvent::Resized` event handler is called. But it doesn't handle HiDPI.

Setting a custom scaling factor doesn't do anything too.

zxsa89675:
Hello, we tried to solve the issue.

This is what we did:

This change adds HiDPI scaling support to the winit::event::WindowEvent::Resized handler in the Application struct. It calculates the new scale factor based on the physical and logical sizes of the window, updates the context's scale factor, and resizes the window with the new scaled dimensions.

You can review changes in this commit: https://github.com/zxsa89675/vizia-vizia-522/commit/a0d1569670fc7d4f56172ed659ab7437d8fc2810.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

geom3trik:
> HiDPI scaling is not working. Actually the source code doc tells, this is handled by the `Resized` event. The relevant code is located in `baseview::WindowEvent::Resized` handler. But this event and even the whole `ApplicationRunner::handle_event` function is never called. Instead the `winit::event::WindowEvent::Resized` event handler is called. But it doesn't handle HiDPI.
> 
> Setting a custom scaling factor doesn't do anything too.

Baseview and winit are two different windowing crates, so you'll only be using one or the other. Are you using vizia to make an audio plugin?

geom3trik:
I just tested this on my device and the vizia gui correctly scales with the scale factor of the system. This is on Windows. Is the gui not scaling for you?

ponchofiesta:
I'm actually not on a computer with HiDPI but will try to create a MVP and a screenshot on the other computer.

ponchofiesta:
I just tried it with current git version of Vizia on my Windows 11 Enterprise 24H2 and two 4K monitors having 175% scaling.

![image](https://github.com/user-attachments/assets/4dae6563-30e5-48aa-887b-c559e850bc54)

```rust
use vizia::prelude::*;

#[derive(Lens)]
struct AppData {}

[...]

PR #543, opened 2025-01-28
Fix init scaling + scaling for scroll

geom3trik:
Sets initial scale factor for windows and updates text when scale factor changes.

**Known Issues**
Scale factor is currently set per application rather than per window as it should be. Fixing this will require fixing layout so that it only runs per window rather than on the whole application. This itself will require a new iterator that skips sub-windows. This will be done in a follow-up PR.


geom3trik:
Fixes #522 
Fixes #534

PR #562, opened 2025-03-26
Add undecorated shadow feature

sout233:
The feature itself comes from winit, and this ensures that its usage is clearly explained within vizia_winit.


geom3trik:
Did you mean to add the IME part to this PR? It looks unfinished. Also does this settings apply to all operating systems or is it Windows only?

sout233:
Sorry, it looks like I accidentally committed extra changes to this branch. That was a mistake.

I’m working on fixing the IME issue in another branch, so I need to resubmit this PR.

PR #564, opened 2025-03-27
 Add undecorated shadow feature

sout233:
The feature itself comes from winit, and this ensures that its usage is clearly explained within vizia_winit.

Re-PR #562


geom3trik:
Unfortunately this feature comes from the `WindowAttributesExtWindow` trait in winit which is Windows only. Therefore adding this method to the window attributes would prevent compilation on other operating systems. You should be able to apply this method directly to the window using the `modify_window` function on `EventContext` in vizia.

sout233:
Okay, I apologize for my negligence. I will close this PR

PR #566, opened 2025-04-04
Add basic support for IME

sout233:
Tested on Windows 10 and 11 with Microsoft IME (Chinese and Japanese).  

There might be some issues with the Japanese IME, but I’m not familiar with how it works, so I’m not sure if it functions correctly.  

Currently, the IME window does not follow the cursor movement—it is only positioned at the textbox. I will address this issue in a separate PR.

### DIFF
--- a/crates/vizia_core/src/context/event.rs
+++ b/crates/vizia_core/src/context/event.rs
@@ -13,7 +13,7 @@ use crate::events::{TimedEvent, TimedEventHandle, TimerState, ViewHandler};
 use crate::prelude::*;
 use crate::resource::ResourceManager;
 use crate::tree::{focus_backward, focus_forward, is_navigatable};
-use vizia_input::MouseState;
+use vizia_input::{ImeState, MouseState};
 
 use skia_safe::Matrix;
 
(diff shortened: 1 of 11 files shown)


geom3trik (approved):
Thank you very much for this! Would you mind telling me how I can test this? I'm on windows 11.

sout233:
Sure!
You can go to **Settings > Time & Language > Language & Region** in Windows and add a language that uses an IME, such as **Chinese or Japanese**.

Here’s the general testing process I usually follow (taking the **Chinese IME** as an example):

1. Run an example app in Vizia that contains a `Textbox` (e.g., the `todo` example).
2. Press **Win + Space** to switch the input method to Chinese.
3. Click the textbox to focus it, and make sure the language indicator in the bottom-right of the taskbar changes from “英” to “中” (or use **Ctrl + Space** to toggle between English and Chinese mode).
4. Type `nihao` — the textbox should show the preedit string **“ni'hao”**, and the IME candidate window should appear, positioned near the textbox. The first candidate should be **“你好”**.
5. Press the **left arrow key** twice to move the cursor between "ni" and "hao", then type `men`. The preedit should now be **“ni'men'hao”**.
6. The first candidate should now be **“你们好”**. Press **space** to confirm — the final committed text in the textbox should be “你们好”.
7. You can repeat the steps with any candidate and it should be correctly committed and replace the previous preedit text.

sout233:
https://github.com/user-attachments/assets/c5f3b4cc-7b91-4ad2-88a0-ba3fe04a8ca3

Issue #575, opened 2025-04-20
Slider crashes on wayland


0HyperCube:
I'm on commit a6fad3251aca138ff2482fc54224dd7b9a96783.

Running `cargo run --example slider` causes an immediate crash:
```
thread 'main' panicked at crates/vizia_core/src/systems/draw.rs:35:42:
called `Option::unwrap()` on a `None` value
```

This occurs because the bounds have a NaN value in. It seems that this comes from a division by zero as setting the default value of `slider_data.size` to 1 fixes the crash.

Some println debugging shows that the code that uses the size is run before `WindowEvent::GeometryChanged` is received. The `Application::new` method calls the closure to construct the initial widgets. It is at this point that the NaN value is set.

geom3trik:
> This occurs because the bounds have a NaN value in. It seems that this comes from a division by zero as setting the default value of `slider_data.size` to 1 fixes the crash.

Yes you're totally right that's definitely the cause of the crash. What I don't understand is why I'm not getting the crash myself on Windows. Out of curiosity, does your solution of setting the value of `slider_data.size` to 1 cause a jump in the rendering? I.e. does it initially render at the wrong size before switching to the correct one?

0HyperCube:
The sliders don't render at the wrong size. However the handles are initially offset and the bar on the vertical slider is incorrect.

Screen recording:
https://github.com/user-attachments/assets/8477be48-47a3-469f-8709-54977d35970b

Initial frame:
![Initial frame with bad sliders](https://github.com/user-attachments/assets/67b57f41-df22-4974-9de5-606d069f2c4f)

geom3trik:
Fixed by #576

PR #581, opened 2025-04-27
Ensure child windows have proper dirty bounds

### LINKED ISSUE #580: Multiwindow Rendering Issues
Right now, I'm just playing around with Vizia and familiarizing myself with it. I ended up trying to do some multiwindow stuff, and after taking some inspiration from the repository example, I ended up with code like this:

https://github.com/dvub/vizia_todo/tree/f04401e5b2c35a48694e7894bc415b5f75f88fe2

However, this ends up creating some fairly problematic rendering issues (as I've mentioned in the Discord server). I'm on Arch Linux, running KDE + Wayland. I can share more details if necessary.

0HyperCube:
Fixes #580

New popup windows did not have the dirty bounds set properly. The dirty bounds included only entities within the window, not the window itself. This is significant as the size of the entity inside is not necessarily the same as the window itself. For example in the reproduction case, the label widget was used.

The fix currently adds the window to the parent redraw list as well as the child redraw list. I'm not sure if it is necessary to add it to the parent redraw list?

I also noticed that creating a new popup window will temporarily shift the contents of the parent window
with shift
without shift

I assume this is because the popup is laid out before the position type is set to absolute. However I'm not really sure how to fix this.

### DIFF
--- a/crates/vizia_core/src/context/mod.rs
+++ b/crates/vizia_core/src/context/mod.rs
@@ -279,6 +279,14 @@ impl Context {
             if let Some(window_state) = self.windows.get_mut(&parent_window) {
                 window_state.redraw_list.insert(entity);
             }
+
+            // If a child window needs redrawing, add itself to the redraw list.
+            // This ensures that the entire window is redrawn: https://github.com/vizia/vizia/issues/580
+            if self.tree.is_window(entity) {
+                if let Some(window_state) = self.windows.get_mut(&entity) {
+                    window_state.redraw_list.insert(entity);
+                }
+            }
         }
     }
 


geom3trik:
I don't think it's necessary to add it to the parent redraw list because each window is redrawn individually.

On a separate note, do you have any idea what's going on with the CI?

geom3trik:
Also I'm struggling to replicate the layout shift. Do you have some example code that reproduces this?

0HyperCube:
This was just using https://github.com/dvub/vizia_todo. I think it probably occurs only on wayland, since it appears that the event ordering is different there (causing #575 to only occur on wayland).

geom3trik (approved):

geom3trik:
> > I don't think it's necessary to add it to the parent redraw list because each window is redrawn individually.
> 
> I've updated this.
> 
> > On a separate note, do you have any idea what's going on with the CI?
> 
> I've added `libwayland-dev` to the ubuntu dependencies which appears to have fix this.

tyvm!

dvub:
I'm on Arch (btw) - is there a dependency that I need to install on my machine for this to work properly?

0HyperCube:
@dvub this PR does not add additional requirements, it just fixes the CI builder. If you cannot compile vizia, you might need some of these dependencies:
apt:
```
sudo apt install libwayland-dev libxkbcommon-dev xorg-dev libx11-xcb-dev libxcb-dri2-0-dev libgl1-mesa-dev libxcb-icccm4-dev libxcursor-dev libxcb-{render,shape,xfixes}0-dev
```
dnf:
```
sudo dnf install wayland-devel libxkbcommon-devel libXcursor-devel libXi-devel libXinerama-devel libXrandr-devel
```

PR #624, opened 2026-03-30
Fix TextBox accessibility

DataTriny:
Populates properties on text runs again, correctly represent multiline textboxes, properly expose placeholder text.

Not sure whether placeholder should be a new style property or not.

Only one weird bug I have attempted to track without luck: sometimes when deleting text using the backspace key at the very end of a textbox, my screen reader does not report the deleted character. Occurrence appears random, potentially more frequent on debug builds.

### DIFF
--- a/crates/vizia_core/src/context/access.rs
+++ b/crates/vizia_core/src/context/access.rs
@@ -97,6 +97,11 @@ impl AccessNode {
         self.node_builder.set_word_lengths(word_lengths);
     }
 
+    /// Sets the placeholder of a text input node.
+    pub fn set_placeholder(&mut self, placeholder: impl Into<Box<str>>) {
+        self.node_builder.set_placeholder(placeholder);
+    }
+
     /// Sets the step for a numerical node.
     pub fn set_numeric_value_step(&mut self, value: f64) {
         self.node_builder.set_numeric_value_step(value);
--- a/crates/vizia_core/src/systems/accessibility.rs
+++ b/crates/vizia_core/src/systems/accessibility.rs
@@ -57,8 +57,6 @@ pub fn accessibility_system(cx: &mut Context) {
                     },
                 }));
             }
-
-            // }
         }
 
         cx.style.reaccess.clear();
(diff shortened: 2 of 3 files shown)


geom3trik (approved):

DataTriny:
I have tracked a bug on our side while investigating this. While I have strong reasons to think  two bugs are linked, I have not tested the fix here since I have not attempted to bump the AccessKit version. Vizia is the only implementation where I have seen text change or caret updates being missed.

geom3trik:
> I have tracked a bug on our side while investigating this. While I have strong reasons to think two bugs are linked, I have not tested the fix here since I have not attempted to bump the AccessKit version. Vizia is the only implementation where I have seen text change or caret updates being missed.

If you did want to test out any potential fix I've updated the accesskit dependency on the accesskit2 branch and attempted to rebase your changes to the textbox. This branch builds on top of a signals rework though #619 so if it’s broken it might be because of that.

DataTriny:
Seems you're still missing changes  from `Node::word_lengths` to `Node::word_starts` in your branch. The catch is that this property limits text runs to 255 characters in length. Therefore you need to provide `Node::next_on_line` and `Node::previous_on_line` on text runs that are part of the same line.

geom3trik:
> Therefore you need to provide `Node::next_on_line` and `Node::previous_on_line` on text runs that are part of the same line.

How do these work sorry? I checked the docs but there seems to be no info on these methods.

DataTriny:
Sorry for the lack of documentation... If a text run is followed by another text run that is part of the same line, then you must set `next_on_line` to the ID of that following run. Similarly, if a text run is preceeded by another text run that is part of the same line, then you must set `previous_on_line` to the ID of that preceeding text run.

DataTriny:
As for the character deletion bug, as I described |here](https://github.com/AccessKit/accesskit/pull/704) it sometimes happen when deleting the last character so that the last text run is removed, such as when a long line would previously span two lines but can fit on one after deletion. I was actually able to reproduce the bug somewhere else, so not something Vizia specific.

PR #628, opened 2026-04-05
Make Knob accessible

DataTriny:
Makes the Knob view accessible by giving it a role, exposing its current, minimum and maximum value, and have it respond to Increment, Decrement and SetValue actions. Forcing percentage values is not ideal though, it would be better to report the real value, like decibel level.

I don't think any accessibility API has a dedicated role for Knob so AccessKit does not have one either, you might want to set `Node::role_description` once you have localization in place, to provide a better user experience.


geom3trik:
Thanks for this. We actually do have localization in place already so I’ll look at adding role description.

geom3trik (approved):

PR #632, opened 2026-04-18
style: fix clippy warnings to unblock CI

paravozz:
The Test workflow runs `cargo clippy` with `RUSTFLAGS: '-D warnings'`, which turns clippy warnings into errors. `main` currently has 41 such warnings that fail CI on every PR.

This PR addresses them all:

- `collapsible_match`: rewrite nested `if` blocks inside `match` arms as match guards. Auto-applied by `cargo clippy --fix` for most cases; four remaining instances in `views/menu.rs` and `views/popup.rs` adjusted manually due to control-flow nuance.
- `unnecessary_cast` in `vizia_winit/src/application.rs`: drop two `as i32` casts where the operand is already `i32`.
- `needless_range_loop` in `text/editable_text.rs::next_line_break`: replace manual index counter with `(from..).zip(bytes)`.

All fixes are mechanical — no behaviour change. Workspace reformatted with `cargo fmt --all` to satisfy the `Check Format` step.

### DIFF
--- a/crates/vizia_core/src/events/event_manager.rs
+++ b/crates/vizia_core/src/events/event_manager.rs
@@ -208,33 +208,31 @@ fn internal_state_updates(cx: &mut Context, window_event: &WindowEvent, meta: &m
             cx.drop_data = Some(drop_data.clone());
         }
 
-        WindowEvent::MouseMove(x, y) => {
-            if !x.is_nan() && !y.is_nan() {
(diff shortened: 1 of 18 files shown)


geom3trik:
Thank you for the effort but I’m holding off on this until the widgets-update PR #627 is merged so I don’t have issues rebasing.

paravozz:
Seems like since #627 is merged and #633 already fixed all warnings, I'm closing this one.

Issue #639, opened 2026-04-20
`ResourceManager::current_translation` fallback dead-ends at empty default bundle, ignoring other registered translations


paravozz:
## Summary

When `ResourceManager::current_translation` is asked for a locale that has no registered translation, the fallback path returns the empty default bundle that `ResourceManager::new` seeds into the `translations` map — even when other real translations (e.g. `en-US`) are registered. Messages present in the registered translation are therefore unreachable via the fallback.

Observable as a failing test already in the tree:

```rust
// crates/vizia_core/src/resource/mod.rs
#[test]
fn current_translation_uses_default_bundle_when_requested_locale_missing() {
    let mut manager = ResourceManager::new();
    manager.add_translation("en-US".parse().unwrap(), "hello = Hello".to_string()).unwrap();
    let bundle = manager.current_translation(&"zz-ZZ".parse().unwrap());
    let message = bundle.get_message("hello");
    assert!(message.is_some());  // currently fires
}
```

Reproducible on `main` at `12ff2584` via `cargo test -p vizia_core --lib current_translation_uses_default_bundle_when_requested_locale_missing`. The CI workflow only runs `cargo fmt --check` + `cargo clippy`, not `cargo test -p vizia_core`, which is why this has been silent.

## Diagnosis

1. `ResourceManager::new` seeds `translations` with `{ LanguageIdentifier::default(): empty_bundle }` (line ~295).
2. `add_translation("en-US", …)` inserts an `en-US` bundle alongside it.
3. `current_translation(&"zz-ZZ")` calls `translation_locales(&"zz-ZZ").into_iter().next()`.
4. `translation_locales` calls `negotiate_translation_locale("zz-ZZ")`, which:
    - Finds no `zz-ZZ` key directly.
    - Collects `available = [en-US]` (filters out default).
    - Calls `fluent_langneg::negotiate_languages(&["zz-ZZ"], ["en-US"], Some(&LanguageIdentifier::default()), Filtering)`.
    - No match, fluent-langneg returns the passed-in default → `LanguageIdentifier::default()`.
5. `current_translation` grabs that locale from the map — the empty seed bundle — and returns it. `get_message("hello")` is None.

So the `en-US` bundle is **right there in the map** but the negotiation deliberately routes fallback to the empty default seed rather than any real translation.

## Proposed directions

Three shapes — flagging as a design call because "default bundle" semantics aren't pinned down in code:

### 1. Fall back to `self.language` instead of `LanguageIdentifier::default()`

[...]

PR #641, opened 2026-04-20
fix(resource): fall back to registered bundle when requested locale is missing

paravozz:
## Problem

See #639 for the full diagnosis. Summary: `ResourceManager::current_translation` for a locale that has no registered translation routes through a fallback that targets `LanguageIdentifier::default()` — the empty seed bundle constructed in `ResourceManager::new`. So misses land there instead of on any registered translation, and `get_message` returns `None` even when a real translation exists right next to the seed in the map.

## Fix

In `negotiate_translation_locale`, build the fluent-langneg fallback from the registered translations themselves:

1. Prefer `self.language` if it happens to be one of the registered locales (common case when the user's system locale matches a registered translation, after `add_translation` runs `renegotiate_language`).
2. Otherwise pick the first registered translation. This covers the case where `renegotiate_language` couldn't match `sys_locale` against any registered locale and left `self.language` at the empty default — typical on systems whose locale doesn't match the registered translations (e.g. `ru-RU` system, only `en-US` registered).

The early-return for "only the default seed is registered" is unchanged — that path still returns the empty default without panicking.

This is direction (1) from the issue body, which you endorsed.

## Test plan

- [x] Renames the existing failing regression test (`current_translation_uses_default_bundle_when_requested_locale_missing`) to accurately describe the new behaviour — it now passes.
- [x] Adds `current_translation_returns_registered_bundle_for_exact_match` as a forward-direction sanity check (registered exact-match returns its own bundle).
- [x] Adds `current_translation_returns_empty_default_when_no_translations_registered` to pin the "seed-only" edge case — that path is unchanged and still returns the empty default.
- [x] `cargo test -p vizia_core --lib resource::` — 7/7 pass.
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy` clean (CI-equivalent).

### DIFF
--- a/crates/vizia_core/src/resource/mod.rs
+++ b/crates/vizia_core/src/resource/mod.rs
@@ -346,11 +346,8 @@ impl ResourceManager {
         // is one of them, otherwise the first registered translation. `available` is
         // non-empty here (checked above), so `available.first()` is always `Some`.
         let first_available = *available.first().expect("non-empty checked above");
-        let fallback = if available.iter().any(|&l| l == &self.language) {
-            &self.language
-        } else {
-            first_available
-        };
+        let fallback =
+            if available.contains(&&self.language) { &self.language } else { first_available };
         let langs = fluent_langneg::negotiate::negotiate_languages(
             &[locale],
             &available,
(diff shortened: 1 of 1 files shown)


geom3trik (approved):

geom3trik:
Just a clippy warning to fix. Once this is merged I'll add testing to CI.

paravozz:
@geom3trik fixing it

PR #645, opened 2026-04-21
Fix scrollbar offset not round-tripping through window minimize/restore.

Greatness7:
Minimizing then restoring the window would cause scrollbars to drift. Easy to see in the widget gallery example by scrolling the widget list down slightly and then repeatedly minimizing/restoring the window.

The bug was seemingly caused by these `container_width != 0.0 && container_height != 0.0` guards. The scroll position state would update when minimizing (i.e. before becoming size zero), but the guards prevent subsequent updates when restoring later (from size zero). After removing the guard the minimizing/restoring triggers should now be properly symmetrical.


geom3trik (approved):
I think this is fine but I do wonder if there was a reason for these guards.

PR #647, opened 2026-04-22
Fix `display` modifier bugs.

Greatness7:
Bug 1:

When an element changes state from `Display::None` to `Display::Flex`, we were setting `RELAYOUT` and `REDRAW` flags, but not `REFLOW`, which is needed to tell the text layout system to rebuild internal paragraphs.

Bug 2:

When a previously visible entity becomes `Display::None`, the draw system was skipping it entirely without marking its previous area as dirty, leaving the last rendered frame of the label stuck on the screen.

---

Small standalone example to reproduce the bugs:

```rs
use vizia::prelude::*;

fn main() -> Result<(), ApplicationError> {
    Application::new(|cx| {
        let show_label = Signal::new(false);
        VStack::new(cx, move |cx| {
            Button::new(cx, |cx| Label::new(cx, "Toggle Label Display"))
                .on_press(move |_| show_label.update(|show| *show = !*show));
            // This label starts with Display::None.
            // Without the fix, when toggled to Display::Flex, vizia would not
            // force the REFLOW system flag. So the text engine wouldn't build
            // the paragraph for the label, causing it to render as empty space
            // until a global reflow was forced (such as minimize/maximize).
            // The second bug would only manifest after fixing that one. When
            // pressing the button to re-hide the label, it would remain due to
            // the space it occupied not being detected as dirty by the draw system.
            Label::new(cx, "Hello, world!").display(show_label);
        });
    })
    .title("Display None -> Flex Bug Repro")
    .run()
}
```

---

Updating `draw_system` to fix this ended up with identical rect-union code being repeated in 3 different places so I went ahead and added a small helper function for code reuse. Also added missing `RELAYOUT` handling to `set_system_flags`, not certain it is necessary but it missing looks unintended to me.

### DIFF
--- a/crates/vizia_core/src/context/mod.rs
+++ b/crates/vizia_core/src/context/mod.rs
@@ -323,6 +323,10 @@ impl Context {
     }
 
     pub(crate) fn set_system_flags(&mut self, entity: Entity, system_flags: SystemFlags) {
+        if system_flags.contains(SystemFlags::RELAYOUT) {
+            self.needs_relayout();
+        }
+
         if system_flags.contains(SystemFlags::RESTYLE) {
             self.needs_restyle(entity);
         }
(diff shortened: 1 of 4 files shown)


geom3trik (approved):
Good catch!

PR #659, opened 2026-04-23
fix(vizia_winit): Fix rendering glitches when resizing

marc2332:


geom3trik (commented):
Not really sure I understand this change. How does this fix it? By flushing events whilst resizing?

marc2332:
> Not really sure I understand this change. How does this fix it? By flushing events whilst resizing?

Basically it does what it already did in Windows, except L510

PR #665, opened 2026-04-24
fix(core): invalidate focus when the focused entity becomes non-visible

paravozz:
## Summary (revised per @geom3trik's feedback)

Keyboard input dispatch routes `WindowEvent::KeyDown` directly to `cx.focused`, and nothing on the style-change side drops focus when the focused entity (or an ancestor) becomes `Display::None`. The existing cleanup in `Context::remove` only fires when an entity is *removed* from the tree, not when it merely gets hidden via a `.display(...)` modifier or a CSS rule.

## Symptom

Most visible in any toggle-able container (modal dialogs, collapsible panels). A `Textbox` inside a modal that gets hidden via `Display::None` continues to receive keystrokes. In a `vizia-plug`-based plug-in this means the host DAW keeps routing keys to the invisible textbox until the plug-in window is destroyed; the DAW's transport key (Space) and other shortcuts can't reach the host.

## Fix

Add `invalidate_focus_on_hidden` as a step at the end of `style_system`. When `cx.focused` (or any ancestor) is `Display::None`, or when the focused entity is `disabled`, move focus to the previous focusable element in tab order via the existing `focus_backward` helper. If `focus_backward` returns `None`, fall back to `Entity::root` so keystrokes route to the window rather than to an invisible view.

## Differences from the previous approach

Per @geom3trik's review feedback:

- **Location**: moved from `BackendContext::process_focus_invalidation` (a new backend-driven phase) into a step at the end of `style_system`. All backends pick this up automatically; no per-backend wiring needed.
- **Fallback target**: instead of always dropping focus to `Entity::root`, walk to the previous non-hidden focusable element via `focus_backward` (the same primitive vizia uses for Shift+Tab navigation). Falls back to `Entity::root` only when nothing focusable exists earlier in tab order.

### DIFF
--- a/crates/vizia_core/src/systems/style.rs
+++ b/crates/vizia_core/src/systems/style.rs
@@ -1472,4 +1472,58 @@ pub(crate) fn style_system(cx: &mut Context) {
     for entity in redraw_entities {
         cx.needs_redraw(entity);
     }
+
+    invalidate_focus_on_hidden(cx);
+}
+
+/// Re-validates `cx.focused` after style resolution.
+///
+/// `WindowEvent::KeyDown` dispatch in `event_manager` routes to
+/// `cx.focused` regardless of the focused view's visibility. Without
+/// re-validation, an entity that becomes `Display::None` (or whose
+/// ancestor does) via either a CSS rule or the `.display(...)` modifier
+/// keeps focus indefinitely; the existing cleanup in `Context::remove`
+/// only fires when an entity is *removed from the tree*, not when it
+/// merely becomes non-visible. The most user-visible symptom is a
+/// plug-in `Textbox` inside a modal that gets hidden via
+/// `Display::None` — every subsequent host keystroke continues to be
+/// routed to the invisible textbox.
+///
+/// Runs at the end of [`style_system`]: at this point `display` and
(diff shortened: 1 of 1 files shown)


geom3trik:
Thanks for the fix! I'm not sure whether this should be a backend context thing although I agree that it probably shouldn’t live in the style system. Perhaps a new focus system is required but I’ll table that for now and go with your solution. Also note that modals should be conditionally built with a Binding rather than hiding with Display, and then focus would work correctly.

paravozz:
@geom3trik thanks for the review and the pointer!

Will refactor the downstream modals in my plugin to use conditional Binding too, since now I know the proper way :)

paravozz:
@geom3trik, is there something missing in the PR to be merged?

geom3trik:
> @geom3trik, is there something missing in the PR to be merged?

Apologies. The more I think about this the more I think that this should actually be handled in the style system and in the Display modifier, but rather than moving focus back to the window it should move focus to the previous non-hidden focusable element.

paravozz:
@geom3trik thanks! Changed the approach per your feedback and updated the PR. If it still doesn't fit, please let me know, and I'll iterate or close and postpone.

geom3trik (approved):

PR #670, opened 2026-04-26
vizia_vulkan renderer for headless rendering needs.

r4nd0m1z3r:
This is implementation of "windowing" backend for vizia that allows to render application into a provided Vulkan image and feeding it window events manually (as demonstrated in basic_vulkan example).

Use case for such thing is embedding vizia into other Vulkan applications (I personally use it to embed it into my game engine).

It mostly inspired from vizia_winit, but I could miss something since I didn't do a deep dive into a vizia architecture. I should probably include some default resources implicitly into application, but when comparing output with vizia_winit I got unexpected issues and decided to leave control to the user, since it is a more low level crate anyways.


geom3trik:
wow thank you for this! I'm currently working on support for metal, dx12, and vulkan backends so I will have to see how this fits in once that lands. I appreciate this is more for embedding in an existing vulkan application.

r4nd0m1z3r:
Yes, it is inspired by how Iced handles embedding, but I was dissatisfied with how iced handled styling, yet it is basically the only library aside from egui that even thinks about embedding use-case.
Your library has both great styling and architecture that is fit for embedding without managing 1000 foreign objects.
So it is currently my library of choice and I may contribute more in future.
I would work on dx12 backend, but I'm not very familiar with it and my main OS is Linux with KDE Wayland.
vizia_vulkan is rather small and I think its design can be generalized enough to fit any backend supported by skia.

PR #683, opened 2026-05-01
Set default theme based on system preference

sergio-fferreira:


geom3trik (approved):

geom3trik (approved):
Looks good to me! So if I understand correctly the correct way to use it is to set up some kind of worker thread to periodically poll and check the theme?

sergio-fferreira:
I believe so, in theory Vizia should use `mundy` to subscribe/stream changes to the system theme at real-time.
This PR, however, is strictly for setting the default theme at boot, based on the system's current theme.

PR #693, opened 2026-05-05
refactor(todo): simplify list rendering and state management

### LINKED ISSUE #689: bug: todo example panic when delete todo
When delete a todo, the app panic "index out of bounds" on this line

https://github.com/vizia/vizia/blob/ced64a15181c0f9a51baae145c8d19ff516147aa/examples/todo/src/main.rs#L182

TheAwiteb:
Fixes: https://github.com/vizia/vizia/issues/689

- Remove unused `TodoEvent::ToggleAll` event and `TodoItemModel`
- Replace index-based `filter_todos` function with reactive memo passed directly to `List`
- Update `RemoveTodo` to pass todo index directly to main model
- Use `Signal::map` instead of memo for filter button active states

- Change `.todo-footer_text` color from white to black for footer text visibility

https://github.com/user-attachments/assets/a2bf2583-885b-44b0-a94d-b7dae5225822

### DIFF
--- a/examples/todo/src/main.rs
+++ b/examples/todo/src/main.rs
@@ -1,9 +1,25 @@
 use vizia::{icons::ICON_TRASH, prelude::*};
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 pub struct Todo {
     title: String,
     completed: bool,
+}
+
+impl Todo {
+    /// Create a new todo
+    pub fn new(title: String) -> Self {
+        Self { title, completed: false }
+    }
+
+    /// Determines whether this todo matches the given filter criteria
+    pub fn match_filter(&self, filter: TodoFilter) -> bool {
+        match filter {
+            TodoFilter::All => true,
+            TodoFilter::Completed => self.completed,
+            TodoFilter::Active => !self.completed,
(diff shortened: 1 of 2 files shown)


examples/todo/src/main.rs
```diff
@@ -147,17 +116,17 @@ fn main() -> Result<(), ApplicationError> {
     Application::new(|cx| {
         cx.add_stylesheet(include_style!("src/style.css")).expect("failed to load style");
 
-        let app = TodoApp::new();
-        let text = app.text;
-        let filter = app.filter;
-        let todos = app.todos;
-        let indices = app.indices;
+        let app @ TodoApp { text, filter, todos } = TodoApp::new();
```
geom3trik, line 119: woah, didn’t know you could do this in Rust

geom3trik (approved):
This is great. Much cleaner. Thank you!

examples/todo/src/main.rs
TheAwiteb: Yeah, pattern are powerful in Rust, you can use the same syntax in if-let and `match`, they are all pattern :)

PR #701, opened 2026-05-07
feat(baseview): live user-scale via WindowEvent::SetUserScale

paravozz:
Adds a `WindowEvent::SetUserScale(f64)` variant in vizia_core and a matching handler in vizia_baseview so the user scale factor can change live at runtime, not only at `Application` construction. The runner captures the event during the per-frame flush and resizes the embedded baseview window, bumps `style.dpi_factor` (so every `Pixels(N)` modifier reflows at layout time), updates the root viewport bounds, and recreates the Skia GL surface at the new physical dimensions in lockstep. The block at `vizia_baseview/src/application.rs:262-329` that hinted at this work is commented out and references a non-existent `cx.user_scale_factor()`; this PR fills that gap with a real, reachable path. Tested on macOS in REAPER (CLAP + VST3) with a plugin that emits `SetUserScale` + `request_resize` on a header click; the window and every `Pixels`-driven layout reflows at the new scale without needing to close and reopen the editor.

### DIFF
--- a/crates/vizia_baseview/src/application.rs
+++ b/crates/vizia_baseview/src/application.rs
@@ -239,10 +239,7 @@ impl ApplicationRunner {
         let base_logical_h = self.window_description.inner_size.height as f64;
         let scaled_logical_w = base_logical_w * new_user_scale;
         let scaled_logical_h = base_logical_h * new_user_scale;
-        window.resize(baseview::Size {
-            width: scaled_logical_w,
-            height: scaled_logical_h,
-        });
+        window.resize(baseview::Size { width: scaled_logical_w, height: scaled_logical_h });
 
         let new_dpi_factor = self.window_scale_factor * new_user_scale;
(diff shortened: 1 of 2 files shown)


geom3trik:
How difficult would it be to add this ability to the winit backend too?

paravozz:
@geom3trik winit ignores `user_scale_factor` today, the field is set on `WindowDescription` but nothing in `vizia_winit/src/` reads it (`style.dpi_factor` comes from `winit_window.scale_factor()` alone).

So two places: at construction, then react to `SetUserScale` at runtime via `request_inner_size(base * new_user_scale)` so winit re-fires `Resized` (which already handles surface/bounds reset).

Maybe 30-50 LOC across `application.rs` and `window.rs`. Happy to extend this PR, or land baseview-only and do winit as a follow-up.

geom3trik (approved):

geom3trik:
We can do it as a follow-up at some point.

Issue #698, opened 2026-05-05
Compile error when disable `x11` and enable `wayland` and `clipboard`


TheAwiteb:
I'm not a GUI guy, but as I understand, wayland manage the clipboard per display/window or something like this, so you need to provide the display somehow.

---

Error:

<details>
<summary>Expand</summary>

```
error[E0432]: unresolved import `copypasta::ClipboardContext`
  --> /urmom/vizia/crates/vizia_core/src/context/mod.rs:28:17
   |
28 | use copypasta::{ClipboardContext, ClipboardProvider, nop_clipboard::NopClipboardContext};
   |                 ^^^^^^^^^^^^^^^^ no `ClipboardContext` in the root
   |
note: found an item that was configured out
  --> /urmom/copypasta-0.10.2/src/lib.rs:60:10
   |
59 | #[cfg(feature = "x11")]
   |       --------------- the item is gated behind the `x11` feature
60 | pub type ClipboardContext = x11_clipboard::X11ClipboardContext;
   |          ^^^^^^^^^^^^^^^^
note: found an item that was configured out
  --> /urmom/copypasta-0.10.2/src/lib.rs:62:10
   |
61 | #[cfg(windows)]
   |       ------- the item is gated here
62 | pub type ClipboardContext = windows_clipboard::WindowsClipboardContext;
   |          ^^^^^^^^^^^^^^^^
note: found an item that was configured out
  --> /urmom/copypasta-0.10.2/src/lib.rs:64:10
   |
63 | #[cfg(target_os = "macos")]
   |       ------------------- the item is gated behind the `macos` feature
64 | pub type ClipboardContext = osx_clipboard::OSXClipboardContext;
   |          ^^^^^^^^^^^^^^^^
note: found an item that was configured out
  --> /urmom/copypasta-0.10.2/src/lib.rs:66:10
   |
65 | #[cfg(target_os = "android")]
   |       --------------------- the item is gated behind the `android` feature
66 | pub type ClipboardContext = nop_clipboard::NopClipboardContext; // TODO: implement AndroidClipboardContext
   |          ^^^^^^^^^^^^^^^^
note: found an item that was configured out
  --> /urmom/copypasta-0.10.2/src/lib.rs:68:10
   |
67 | #[cfg(target_os = "ios")]
   |       ----------------- the item is gated behind the `ios` feature
68 | pub type ClipboardContext = nop_clipboard::NopClipboardContext; // TODO: implement IOSClipboardContext
[...]

PR #711, opened 2026-05-18
Add Wayland clipboard support and provider selection

geom3trik:
Introduce a platform-aware clipboard provider selection and Wayland initialization. Adds default_clipboard_provider() in vizia_core to choose a real ClipboardContext when available or fall back to NopClipboardContext; adjusts copypasta imports with cfgs. In vizia_winit, import copypasta's Wayland helper and winit raw display types under appropriate cfgs, add init_wayland_clipboard() to create clipboards from the window's Wayland display handle, and call it after creating the main window. These changes enable using an external Wayland clipboard when running on Wayland-only targets (with the clipboard feature) and ensure a safe fallback to a no-op clipboard when necessary.

### DIFF
--- a/crates/vizia_core/src/context/mod.rs
+++ b/crates/vizia_core/src/context/mod.rs
@@ -25,8 +25,22 @@ use std::{
 use vizia_id::IdManager;
 use vizia_window::WindowDescription;
 
+#[cfg(all(
+    feature = "clipboard",
+    not(all(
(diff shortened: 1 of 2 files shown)


sergio-fferreira:
Seems to be working on Linux+Wayland

geom3trik:
Fixes #698

PR #749, opened 2026-07-05
Fix infinite loop in Context::modify_timer (#530)

### LINKED ISSUE #530: Creating two timers, where the first timer never finishes, enters an infinite loop
Debugged as so:

Context::running_timers containing:
- Timer with ID 1, with interval 1s

We then create a second timer, also with interval 1s

Upon adding the second timer, we push it to the `running_timers` vector, then call `modify_timer` which then enters a while loop... on the top of a binary heap waiting for the next timer to change?
But this is never going to happen, since the first timer infinite loops. Plus, since we're single threaded, even if it did expire... we wouldn't actually do the expiring, since we're stuck in this infinite loop.

Personally, I'm not quite sure what modify_timer() is trying to do at all. Is it even needed if the timer isn't running?

```rust
    pub fn start_timer(&mut self, timer: Timer) {
        let current = self.current;
        if !self.timer_is_running(timer) {
            let timer_state = self.timers[timer.0].clone();
            self.running_timers.push(timer_state);
        }

        self.modify_timer(timer, |timer_state| {
            ...
        });
    }

    pub fn modify_timer(&mut self, timer: Timer, timer_function: impl Fn(&mut TimerState)) {
		// Gets stuck here
        while let Some(next_timer_state) = self.running_timers.peek() {
            if next_timer_state.id == timer {
                let mut timer_state = self.running_timers.pop().unwrap();

                (timer_function)(&mut timer_state);

                self.running_timers.push(timer_state);

                return;
            }
        }

		...
    }
```

lxndrbe:
## Summary

Fixes #530.

`Context::modify_timer` searched `running_timers` (a min-heap ordered by trigger time, not by timer id) by only ever calling `peek()` on the top entry:

```rust
while let Some(next_timer_state) = self.running_timers.peek() {
    if next_timer_state.id == timer {
        // ... pop, modify, push, return
    }
}
```

If the target `timer` wasn't the current top-of-heap entry, the `if` never matched, nothing was popped, and the next loop iteration's `peek()` returned the exact same element again - an infinite loop. Since this runs on the UI thread with no blocking wait involved, it presents as a hard hang rather than a panic or a visible deadlock in a debugger's thread-wait view.

This is easy to trigger with completely ordinary usage: any app with two timers running at once (e.g. the built-in `Environment::caret_timer`, registered for every `Context` and started on a `Textbox` focus-in event, racing an application-level timer started via `cx.add_timer`/`cx.start_timer`) can hit it as soon as the two aren't in the "lucky" heap order.

## Fix

Pop non-matching entries aside into a `Vec` while searching, keep peeking the new top, and once found (or if the heap is exhausted), push everything back - including the modified target - before returning. Same shape as a normal linear/heap walk that needs to temporarily displace elements it's not looking for.

## Testing

- `cargo check -p vizia_core` passes.
- Verified against a real repro in a downstream CLAP audio plugin (two timers: a custom app timer + `Environment::caret_timer` after a `Textbox` gained focus) - hung reliably before this fix, opens and runs normally after.

No behavior change for the already-working path (target found at heap top on the first `peek()`).

### DIFF
--- a/crates/vizia_core/src/context/mod.rs
+++ b/crates/vizia_core/src/context/mod.rs
@@ -866,6 +866,14 @@ impl Context {
 
     /// Modifies the state of an existing timer with the provided `Timer` id.
     pub fn modify_timer(&mut self, timer: Timer, timer_function: impl Fn(&mut TimerState)) {
+        // `running_timers` is a min-heap ordered by trigger time, not by id -
+        // the target timer isn't necessarily at the top. The previous version
+        // only ever `peek()`ed the top and returned early on a mismatch
+        // without advancing, so a target that wasn't already the top-of-heap
+        // entry spun this loop forever (`peek()` keeps returning the same
+        // element). Pop non-matching entries aside, keep searching, and push
+        // everything back (including the modified target) before returning.
+        let mut passed_over = Vec::new();
         while let Some(next_timer_state) = self.running_timers.peek() {
             if next_timer_state.id == timer {
                 let mut timer_state = self.running_timers.pop().unwrap();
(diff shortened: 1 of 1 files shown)


geom3trik:
Thanks for this!

PR #754, opened 2026-07-16
Fix running timer lookup loops

geom3trik:
Timer lookup in `Context` and `EventContext` no longer relies on `peek()` inside a `while` loop, which could loop forever when the top heap entry didn’t match the requested timer ID. The code now searches all running timers (`iter`/`iter_mut`) and, for mutable access, temporarily takes and restores the heap after applying updates. This makes `modify_timer` and `query_timer` reliably find non-top running timers without hanging.


geom3trik:
Fixes #530

geom3trik:
Alternative to closed PR #749

PR #760, opened 2026-07-24
Add track hold steps feature to slider control

ryanroe:
Scenario: when slider used as gain control for an equalizer, if user clicks at the max position, the gain will suddenly goes to the maximum level, which harms.

added a track hold steps feature to prevent this, users can use with:

```rs
Slider::new(cx,value).track_hold_steps(true);
```

the name `track_hold_steps` may looks weird, but i don't know how to name this feature

### DIFF
--- a/crates/vizia_core/src/views/slider.rs
+++ b/crates/vizia_core/src/views/slider.rs
@@ -1,4 +1,5 @@
 use std::ops::Range;
+use std::time::Duration;
 
 use crate::prelude::*;
 use accesskit::ActionData;
@@ -10,6 +11,9 @@ pub(crate) enum SliderEvent {
     SetMin,
     SetMax,
     ResetDefault,
+    StartTrackHold,
+    EndTrackHold,
+    TrackHoldRepeat,
 }
 
 /// The slider control can be used to select from a continuous set of values.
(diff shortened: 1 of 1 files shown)


geom3trik:
I'm sorry but I am not understanding what this feature does. So what happens when you click the max position with this feature turned on?

ryanroe:
The value will increase by 1 if the step of slider is 1.

say current value is 10, max is 100, after click at max position

feature off:  value becomes 100
feature on: value becomes 11

geom3trik:
Ah okay I understand now. We have a similar feature for scroll bars where clicking the track can either scroll by a fixed amount or scroll to the cursor.

sourcebox:
I had a look at several plugins and most of them don't move a slider at all when just clicking on it. They require a slider to be dragged. So overall, I think this feature makes sense.

ryanroe:
> Ah okay I understand now. We have a similar feature for scroll bars where clicking the track can either scroll by a fixed amount or scroll to the cursor.

will you apply that behaviour to slider?

geom3trik:
> will you apply that behaviour to slider?

Well I guess that’s what this PR is doing. Sorry for the delay in reviewing, I should hopefully get to it by the end of the weekend. One part I am not sure about is the repeat timer. I don’t know plug-ins that well, is this a common behavior?

ryanroe:
> > will you apply that behaviour to slider?
> 
> Well I guess that’s what this PR is doing. Sorry for the delay in reviewing, I should hopefully get to it by the end of the weekend. One part I am not sure about is the repeat timer. I don’t know plug-ins that well, is this a common behavior?

- for common behaviour:

if the behaviour means: `click for step, hold for fast move`,  as far as i know, only sliders and scrollbars use this, but i think it can be extended to all range-based control, and we can then further extract this behaviour (like WPF's attached behavior, but i am not sure if this can be done with Rust)



- for repeat timer:
if this means implementation detail in this PR, i use two timers to get this work.
  - delay timer: tells how long user has pressed and held , when the timer stops, starts repeat timer(value fast move)
  - repeat timer: controls how fast it moves to target value, the lower interval it has , the faster it moves to target value.

and sorry but i'm not urging you to review or merge:)

